PR #13, opened 2019-01-20
stack-exec-simplify : remove --interleaved-output from stack exec

shayne-fletcher-da:
This PR adjusts the `stack-exec` commands to not include the `--interleaved-output` option - not necessary in these cases.


neil-da:
LGTM

PR #42, opened 2019-03-08
Adjust to GHC head

shayne-fletcher-da:
In this PR:
  - Switched to a new scheme that ships the `hadrian` generated `Lexer.hs` and `Parser.hs` (and does not ship `Lexer.x` and `Parser.y`);
  - Added code to delete `compiler/Parser.y` and `compiler/Lexer.x` after `hadrian` has consumed them (required by the scheme);
  - Adjusted to a change in the definition of `ParseResult` (in `Lexer.x` - `PFailure` now gets only one argument, not three);
  - Changed the GHC repository remote to gitlab and the checkout protocol to https;
  ~~- Added `--flavour=quickest` and `-j` to the `hadrian` flags (probably meaningless but they appear to do no harm);~~
 - Fixed the `stack setup` command in `.travis.yml` to write to `/dev/null` not `nul` (should probably put a `1>&2` in there as well but didn't do it here).


shayne-fletcher-da:
Aside: The below seems to happen more often than we'd like. Contention?
```

fatal: unable to access 'https://gitlab.haskell.org/ghc/ghc.git/': gnutls_handshake() failed: Handshake failed
CI.hs: Failed when running system command: git clone https://gitlab.haskell.org/ghc/ghc.git --recursive
```

.travis.yml
```diff
@@ -7,6 +7,6 @@ os:
 script:
 - curl -sSL https://raw.github.com/ndmitchell/hlint/master/misc/travis.sh | sh -s .
 - curl -sSL https://get.haskellstack.org/ | sh
-- stack setup > nul
+- stack setup > /dev/null
```
neil-da: Cool - but note that doing `>&2` or whatever to lose stderr would be wrong. We only suppress stdout because it generates a huge amount of pointless output (one line per unzip'd file). stderr is still super valuable.

shayne-fletcher-da:
I'll do follow up PRs. For now, let's recover master and land this thing.

PR #44, opened 2019-03-09
Get errors

shayne-fletcher-da:
In this PR, `mini-hlint`'s error handling code is adapted to upstream changes to `data ParseResult`.

### DIFF
--- a/examples/mini-hlint/src/Main.hs
+++ b/examples/mini-hlint/src/Main.hs
@@ -101,6 +101,5 @@ main = do
         POk _ m -> analyzeModule flags m
         PFailed s ->
           mapM_ putStrLn $
-          fmap (showSDoc flags) $
-          pprErrMsgBagWithLoc (snd $ getMessages s flags)
+          showSDoc flags <$> pprErrMsgBagWithLoc (snd $ getMessages s flags)
     _ -> fail "Exactly one file argument required"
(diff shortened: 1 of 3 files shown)


examples/mini-hlint/src/Main.hs
```diff
@@ -99,7 +99,9 @@ main = do
       let flags = defaultDynFlags fakeSettings fakeLlvmConfig
       case parse file flags s of
         POk _ m -> analyzeModule flags m
-        PFailed _ ->
-          -- putStrLn $ showSDoc flags $ pprLocErrMsg $ mkPlainErrMsg flags loc err
-          putStrLn "Fix me" -- to do; call `getErrorMessages` to replace the old code above
+        PFailed s -> sequence_
+          [ putStrLn msg
+          | msg <- showSDoc flags <$>
```
neil-da: I'd have moved the `showSDoc` so it read `[ putStrLn $ showSDoc flags msg | msg <- ...]`. The use of `<$>` as map is unusual.

shayne-fletcher-da: Yes, that's nice too. That said, are you ok with keeping the current phrasing?

shayne-fletcher-da: Going with the following which eliminates the `map`.
```
PFailed s -> sequence_
          [ putStrLn (showSDoc flags msg)
          | msg <- pprErrMsgBagWithLoc (snd $ getMessages s flags)
          ]
```

PR #50, opened 2019-03-31
Introduce new package ghc-lib-parser

shayne-fletcher-da:
In this PR, `ghc-lib` has been refactored into two packages : `ghc-lib-parser` and `ghc-lib` with the latter depending on the former.


CI.hs
```diff
@@ -13,34 +13,34 @@ import System.Time.Extra
...
-
-    -- Generate an sdist and extract it to ensure it works.
+    cmd "stack exec -- ghc-lib-gen ghc --ghc-lib-parser"
     stackYaml <- readFile' "stack.yaml"
     writeFile "stack.yaml" $ stackYaml ++ unlines ["- ghc"]
-    cmd "stack sdist ghc --tar-dir=."
-    [tarball] <- filter (isExtensionOf ".tar.gz") <$> getDirectoryContents "."
-    cmd $ "tar -xf " ++ tarball
+    tarball <- sDistCreateExtract
+    renameDirectory (dropExtension $ dropExtension tarball) "ghc-lib-parser"
```
neil-da: `dropExtensions` ?

CI.hs
```diff
@@ -13,34 +13,34 @@ import System.Time.Extra
...
+    removeFile tarball
+    removeFile "ghc/ghc-lib-parser.cabal"
+
+    -- Make and extract an sdist of ghc-lib.
+    cmd "cd ghc && git checkout . && cd .."
```
neil-da:
```suggestion
    cmd "cd ghc && git checkout ."
```

These run in separate shells - changing directory back after is pointless

ghc-lib-gen/src/Main.hs
```diff
@@ -251,7 +457,85 @@ generateCabal = do
...
+    bin <- (:[]) <$> readCabalFile cabalFileBinary
+    let askField from x = nubSort $ concatMap (`askCabalField` x) from
+    let askFiles from x = nubSort $ concatMap (`askCabalFiles` x) from
+
+    writeFile "ghc-lib-parser.cabal" $ unlines $ map trimEnd $
+        -- header
+        ["cabal-version: >=1.22"
+        ,"build-type: Simple"
+        ,"name: ghc-lib-parser"
+        ,"version: 0.1.0"
```
neil-da: I assume we then manually change the version?

shayne-fletcher-da: Yes.

PR #55, opened 2019-04-05
Update azure-pipelines.yml for Azure Pipelines [skip ci]

dbreydo-da:
minimizing CI work: run builds on PRs targeting `master` and on a commit to `master`


dbreydo-da:
> Thanks. What is the semantics of the `batch` flag?

From the documentation: 
```
 If you set batch to true, when a build is running, the system waits until the build is completed, then queues another build of all changes that have not yet been built.
```

PR #61, opened 2019-05-10
Parser error handling improvements

shayne-fletcher-da:
This PR improves error checking in the mini-hlint example.

### DIFF
--- a/examples/mini-hlint/src/Main.hs
+++ b/examples/mini-hlint/src/Main.hs
@@ -102,7 +102,7 @@ main = do
           report flags (snd (getMessages s flags))
         POk s m -> do
           let (warns, errs) = getMessages s flags
-          if (not (null errs))
+          if not (null errs)
             then report flags errs
             else do
               unless (null warns) $ report flags warns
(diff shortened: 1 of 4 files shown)


examples/mini-hlint/src/Main.hs
```diff
@@ -98,9 +98,19 @@ main = do
...
-          | msg <- pprErrMsgBagWithLoc $ snd $ getMessages s flags
-          ]
+        PFailed s ->
+          report flags (snd (getMessages s flags))
+        POk s m -> do
+          let (warns, errs) = getMessages s flags
+          if not (null errs)
+            then report flags errs
+            else do
+              unless (null warns) $ report flags warns
```
neil-da: Why not just `report $ errs ++ warns`? On `[]` report does nothing.

shayne-fletcher-da: (1) We agree we will not proceed to `analyzeModule` if there are (so called "non-fatal") errors.
(2) Nice ideas on reporting but `warns` and `errs` can't be combined that way (not lists). Next iteration though will unconditionally report warnings in the `POk` case though.

PR #66, opened 2019-05-20
Remove references to getOrSetLibHSghc

cocreature:
By default, GHC uses a bunch of calls to getOrSetLibHSghc when
-DSTAGE=2. This resulted in segfaults as described in
https://github.com/ndmitchell/hlint/issues/637. In that case it was
caused by a change in the representation of the FastString table but
even if there is no change in representation, we want to avoid any
interaction between ghc and ghc-lib.

I grepped for all usages of getOrSetLibHSghc and patching them seemed
easier than switching everything to -DSTAGE=1 (e.g. in MachDeps.h we
really want -DSTAGE=2).


neil-da:
Awesome work!

PR #67, opened 2019-05-22
Try to recover the mini-compile example

shayne-fletcher-da:
Provide a definition for record field `sIntegerLibraryType`.


ndmitchell:
👍

PR #70, opened 2019-05-26
New ghc-lib-parser source file 'PlainPanic.hs'

shayne-fletcher-da:
This PR adds a newly added ghc source file to `ghc-lib-parser`.


neil-da:
Nice. How are you detecting these? Does some CI fail or similar?

shayne-fletcher-da:
A combination of carefully monitoring MRs landing/about to land on GHC master,  rebasing immediately when they do, confirming the need to update `ghc-lib-gen` by way of CI failures and then implementing the fixes. I'm happy to say that CI here and in https://github.com/digital-asset/ghc has begun to pay its way.

PR #73, opened 2019-05-30
Parse pragmas

shayne-fletcher-da:
This PR enhances `mini-hlint` by adding code to handle dynamic pragmas. For example, before this change, the program
```haskell
{-# LANGUAGE ImportQualifiedPost #-}
module M where
import Data.List qualified
```
generated the error,
```
Found `qualified' in postpositive position.
```
After this change, the program parses. Tests have been added to check (1) the above statement holds and (2) that exceptions potentially thrown in the implementation in response to malformed pragmas are caught.


examples/mini-hlint/mini-hlint.cabal
```diff
@@ -9,10 +9,12 @@ copyright:           Digital Asset 2018-2019
...
 
+
 executable mini-hlint
   main-is:             Main.hs
   build-depends:       base >=4.11
                      , extra
                      , ghc-lib-parser
   default-language:    Haskell2010
   hs-source-dirs:      src
+  ghc-options: -Wmissing-exported-signatures
```
neil-da: I generally don't put ghc warnings options in the .cabal file. Warnings are for my benefit, while the Cabal is for distribution. Instead I put these warning checks in the .ghci file and in the CI.

examples/mini-hlint/src/Main.hs
```diff
@@ -107,15 +124,20 @@ main = do
...
-          report flags $ snd (getMessages s flags)
-        POk s m -> do
-          let (warns, errs) = getMessages s flags
-          report flags warns
-          report flags errs
-          when (null errs) $ analyzeModule flags m
+      flags <-
+        parsePragmasIntoDynFlags
+          (defaultDynFlags fakeSettings fakeLlvmConfig) file s
+      whenJust flags
```
neil-da: `whenJust flags $ \flags ->` would be the normal way to use `whenJust` here.

ghc-lib-gen/src/Main.hs
```diff
@@ -187,6 +187,7 @@ parserModules =
   , "GhcMonad"
   , "GhcPrelude"
   , "HaddockUtils"
+  , "HeaderInfo"
```
neil-da: Is this change unrelated but just going along at the same time? (Fine with that, just checking its not a mistake that went along for the ride)

shayne-fletcher-da: `HeaderInfo` was in `ghc-lib`. I moved it to `ghc-lib-parser` so that `hlint` is not required to depend on `ghc-lib` but can remain satisfied with `ghc-lib-parser`.

examples/mini-hlint/mini-hlint.cabal
shayne-fletcher-da: That was a mistake - detritus left behind from some debugging. Will fix.

examples/mini-hlint/src/Main.hs
shayne-fletcher-da: Will fix

shayne-fletcher-da: An opportunity for a new `hlint` hint here?

neil-da: Perhaps... but it's more subjective I guess.

PR #75, opened 2019-06-08
Adapt to upstream changes

shayne-fletcher-da:
This PR adds `GHC.BaseDir` to `ghc-lib-parser` introduced in `387050d0e26a9e6466b31c9d8e4e4f6273c64c9e`.

### DIFF
--- a/ghc-lib-gen/src/Main.hs
+++ b/ghc-lib-gen/src/Main.hs
@@ -164,6 +164,7 @@ parserModules =
   , "FiniteMap"
   , "ForeignCall"
   , "GhcNameVersion"
+  , "GHC.BaseDir"
   , "GHC.Exts.Heap"
   , "GHC.Exts.Heap.ClosureTypes"
   , "GHC.Exts.Heap.Closures"


shayne-fletcher-da:
No, this is on `HEAD`.

ndmitchell:
Does adding this extra file cause any problems if/when we generate a ghc-lib from the 8.8 branch?

shayne-fletcher-da:
No. Last week I constructed branch `ghc-lib-8.8.1-alpha1` specialized for 8.8.1 while at the same time containing the recent fixes made by Moritz for C obfuscation and son on. It was the basis of the `ghc-lib-8.8.0.20190424` release that went out on Jun 5.

ndmitchell:
Is the only delta the list of modules? If so, would automating that let us get rid of the branch and stick to one version of ghc-lib?

shayne-fletcher-da:
No, it is not the only delta. Automating the module list for `ghc-lib-parser` (in so far as possible) would close the gap but no, it is not the only difference.

shayne-fletcher-da:
To clarify, the remaining deltas are in the examples `mini-hlint` and `mini-compile` : modules imported, settings logic and parse tree differences (e.g. `PFailed _` vs `PFailed _ _ _`). I suppose these differences might be overcome today with judicious CPP use. It is possible of course that if we went this way it could become yet more "dense" over time. We are agreed that in any event, generating the `ghc-lib-parser` module list is desirable in any case, it's as you know "on the stack" (as is sniffing out build dependencies and their constraints).

PR #76, opened 2019-06-11
Calc parser modules

shayne-fletcher-da:
This PR makes a start on a procedure to calculate the list of modules required for `ghc-lib-parser`.


sniff.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
shayne-fletcher-da: This file is temporary. It's a tool for me to work out the procedure. It will get removed in the fullness of time.

shayne-fletcher-da:
Great, well, in the middle of all of this, I've just discovered that something has landed on ghc HEAD in the last hour that has broken us entirely and is responsible for all these failures and the failure on master. Right now I hold aad6115aed884db9ed47ac602ca1bd3b953ea089 responsible.

shayne-fletcher-da:
Looking at that patch and experimenting, it looks like a build can be recovered by changing the define `GHCI` with `HAVE_INTERNAL_INTERPRETER` and undoing the changes to `hadrian/src/Settings/Packages.hs`.  It's a mess.

shayne-fletcher-da:
I suspect, the reason for the last commit is that `stack` has been upgraded and that's why I'm not able to reproduce locally.

shayne-fletcher-da:
Ah yes! Now that I think of it, I'm sure I heard mention at Zurihac that stack was upgraded during the hackathon and was freaking everybody out :)

shayne-fletcher-da:
@neil-da Assuming this goes green, I think it's ready for review!

ghc-lib-gen/src/Main.hs
```diff
@@ -49,7 +51,36 @@ cabalFileLibraries =
...
+-- adjusting.
+ghcLibParserHsSrcDirs :: [Cabal] -> [FilePath]
+ghcLibParserHsSrcDirs lib =
+  nubSort $
+   [ "ghc-lib/stage0/compiler/build"
+   , "ghc-lib/stage1/compiler/build"] ++
+   map takeDirectory cabalFileLibraries ++
+   askFiles lib "hs-source-dirs:"
+
+-- | The "hs-source-dirs" for 'ghc-lib-parser'. Approximation. Needs
```
cocreature:
```suggestion
-- | The "hs-source-dirs" for 'ghc-lib'. Approximation. Needs
```

ghc-lib-gen/src/Main.hs
```diff
@@ -96,205 +127,52 @@ extraFiles =
...
+        ++ ["ghc-lib/stage0/compiler/build/Parser.hs"]
+  putStrLn "# Generating 'ghc/.parser-depends'..."
+  system_ cmd
 
--- | Stub out a couple of definitions in the ghc-heap library that require CMM features,
---   since Cabal doesn't work well with CMM files.
+  buf <- readFile' ".parser-depends"
+  let depends = filter (not . isPrefixOf "#") (lines buf)
+      depends' = filter (isSuffixOf ".hs") depends
+      srcPaths = map snd (mapMaybe (stripInfix ":") depends')
```
cocreature: This looks like a lot of repetition and duplication and it isn’t entirely clear what you are trying to accomplish with that.
I would suggest that you add a comment with an example of an input path and the final path after all this magic.
Presumably your goal here is to strip away the source dir? If so, should this be based on `ghcLibParserHsSrcDirs`?

PR #78, opened 2019-06-17
Stack upgrade highlights some deficienies in CI.hs

shayne-fletcher-da:
The newest version of `stack` is a little stricter than it was highlighting some abuses we were making in `CI.hs`. This sorts it out.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -29,15 +29,19 @@ main = do
     renameDirectory (dropExtensions tarball) "ghc-lib-parser"
     removeFile tarball
     removeFile "ghc/ghc-lib-parser.cabal"
+    cmd "git checkout stack.yaml"
 
     -- Make and extract an sdist of ghc-lib.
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -29,15 +29,19 @@ main = do
     renameDirectory (dropExtensions tarball) "ghc-lib-parser"
     removeFile tarball
     removeFile "ghc/ghc-lib-parser.cabal"
+    cmd "git checkout stack.yaml"
```
cocreature: It would be nice to add some comments on what exactly is going on here. E.g., why do you need to add `- ghc` to the end, which section is that even in the `stack.yaml`, why do we only need this for some things, …

shayne-fletcher-da: Right. In fact, the program is a kind of formalization of the `README` which might help in the interim. I'll make an issue for it.

PR #82, opened 2019-06-18
A revised approach to "path-stripping"

shayne-fletcher-da:


CI.hs
```diff
@@ -23,9 +23,10 @@ main = do
 
     appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
     cmd "stack exec -- ghc-lib-gen ghc --ghc-lib-parser"
-    stackYaml <- readFile' "stack.yaml"
-    writeFile "stack.yaml" $ stackYaml ++ unlines ["- ghc"]
-    tarball <- sDistCreateExtract
+    -- stackYaml <- readFile' "stack.yaml"
```
cocreature: If this is no longer necessary, can we just remove it?

CI.hs
```diff
@@ -35,15 +36,17 @@ main = do
     cmd "cd ghc && git checkout ."
     appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
     cmd "stack exec -- ghc-lib-gen ghc --ghc-lib"
-    stackYaml <- readFile' "stack.yaml"
-    writeFile "stack.yaml" $ stackYaml ++ unlines ["- ghc"]
-    tarball <- sDistCreateExtract
+    -- stackYaml <- readFile' "stack.yaml"
```
cocreature: Same here

ghc-lib-gen/src/Main.hs
```diff
@@ -65,11 +65,14 @@ ghcLibParserIncludeDirs =
 -- adjusting.
 ghcLibParserHsSrcDirs :: [Cabal] -> [FilePath]
 ghcLibParserHsSrcDirs lib =
-  nubSort $
-   [ "ghc-lib/stage0/compiler/build"
-   , "ghc-lib/stage1/compiler/build"] ++
-   map takeDirectory cabalFileLibraries ++
-   askFiles lib "hs-source-dirs:"
+  -- Sort by length so the longest paths are at the front.
```
cocreature: This comment tells me what we are doing but not why this is necessary.

ghc-lib-gen/src/Main.hs
```diff
@@ -65,11 +65,14 @@ ghcLibParserIncludeDirs =
...
+  sortBy (\s t -> if length s > length t then LT
```
cocreature:
```suggestion
  sortBy (flip (comparing length))
```

ghc-lib-gen/src/Main.hs
```diff
@@ -148,25 +151,31 @@ calcParserModules = do
...
-      srcPaths' = map (replace "ghc-lib/stage0/" "") srcPaths
-      srcPaths'' = map (replace "ghc-lib/stage1/" "") srcPaths'
-      srcPaths''' = map snd (mapMaybe (stripInfix "/") srcPaths'')
-      srcPaths'''' = map snd (mapMaybe (stripInfix "/") srcPaths''')
-      srcPaths''''' = map (replace "build/GHC/" "GHC/") srcPaths''''
-      srcPaths'''''' = map (replace "/" ".") srcPaths'''''
-      srcs = map (dropSuffix ".hs") (filter (isSuffixOf ".hs") srcPaths'''''')
-      -- 'GHCi.FFI doesn't get deduced but is needed
-      -- 'HeaderInfo' because we prefer it here rather than `ghc-lib`
+      srcPaths = map (trim . snd) (mapMaybe (stripInfix ":") depends')
```
cocreature:
```suggestion
      modulePaths = map (trim . snd) (mapMaybe (stripInfix ":") depends')
```

ghc-lib-gen/src/Main.hs
```diff
@@ -148,25 +151,31 @@ calcParserModules = do
...
+      srcDirs = ghcLibParserHsSrcDirs lib
+      srcPaths' = map (replace "/" "") (foldl
```
cocreature:
```suggestion
      strippedModulePaths = map (replace "/" "") (foldl
```

ghc-lib-gen/src/Main.hs
```diff
@@ -148,25 +151,31 @@ calcParserModules = do
...
+        (\acc p -> map (replace (p ++ "/") "") acc)
+        srcPaths
+        srcDirs)
+      srcs = map (dropSuffix ".hs") (filter (isSuffixOf ".hs") srcPaths')
```
cocreature: Are all files `.hs` files? I thought we also had some `.hs-boot` and some `.hsc` files.

ghc-lib-gen/src/Main.hs
cocreature: Shouldn’t we replace `/` by `.` instead to convert from filenames to module names? I would also suggest to do this at the same step where you drop the `.hs` suffix so that you have a single step that converts from filenames to module names.

ghc-lib-gen/src/Main.hs
```diff
@@ -148,25 +151,31 @@ calcParserModules = do
         , "-dep-makefile .parser-depends"
         , "-M"]
         ++ include_dirs
-        ++ ["-ignore-package ghc-lib-parser -package ghc", "-package base"]
-        ++ hs_source_dirs ++ ["-ighc-lib/stage0/libraries/ghc-heap/build"]
+        ++ ["-ignore-package ghc-lib-parser"
+           , "-ignore-package ghc"
+           -- , "-ignore-package ghci"
```
cocreature: If we don’t need that line, let’s remove it.

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: This is the list of exposed-modules we are building. Running `cabal sdist` does indeed pull in `.hs-boot` files but they aren't enumerated here.

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: That was a bug :)

CI.hs
shayne-fletcher-da: ✔️

CI.hs
shayne-fletcher-da: ✔️

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: ✔️

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: Thanks!

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: ✔️

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: ✔️

ghc-lib-gen/src/Main.hs
shayne-fletcher-da: Actually we do but the ghc-lib-parser line we don't. This program is so fiddly to get right (and I'm sure this isn't the end of it...)!

shayne-fletcher-da:
@cocreature Not required but if you have the patience for it, this is ready for another look!

ghc-lib-gen/src/Main.hs
```diff
@@ -65,11 +66,17 @@ ghcLibParserIncludeDirs =
...
-  nubSort $
-   [ "ghc-lib/stage0/compiler/build"
-   , "ghc-lib/stage1/compiler/build"] ++
-   map takeDirectory cabalFileLibraries ++
-   askFiles lib "hs-source-dirs:"
+  -- Sort by length so the longest paths are at the front. We do this
+  -- so that in 'calcParserModules', longer substituions are performed
+  -- before shorter ones (and bad things will happen if that were
+  -- not the case).
+  sortBy (flip (comparing length))
```
neil-da: `sortOn (negate . length)`?

ghc-lib-gen/src/Main.hs
```diff
@@ -65,11 +66,17 @@ ghcLibParserIncludeDirs =
...
+    ([ "ghc-lib/stage0/compiler/build"
+     , "ghc-lib/stage1/compiler/build"
+     , "ghc-lib/stage0/libraries/ghci/build"
+     , "ghc-lib/stage0/libraries/ghc-heap/build"
+     ] ++ map takeDirectory cabalFileLibraries ++
+      askFiles lib "hs-source-dirs:")
```
neil-da: I'd definitely `$` this whole block rather than having surrounding brackets that mess up indentation etc.

PR #84, opened 2019-06-21
Use a $ to reduce syntactic noise

shayne-fletcher-da:
Acting on @neil-da 's suggestion for a $ to reduce parentheses.


shayne-fletcher-da:
I think it's clever but the `flip (comparing length)` seems weakly (subjectively) clearer to me. If you feel strongly that it's better I won't object and make the change :) Let me know!

neil-da:
I find the use of `comparing` more confusing in most cases, but beauty is in the eye of the author, happy to stick with that if you like it.

PR #87, opened 2019-07-03
Advance head

shayne-fletcher-da:
We have now got a fix for adapting to `39f50bff3ea913a7f4b1d915660bcf77b9327e2e`. Let's now see how far we can push this.


CI.hs
```diff
@@ -1,7 +1,8 @@
--- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
--- SPDX-License-Identifier: (Apache-2.0 OR BSD-3-Clause)
+-- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its
+-- affiliates. All rights reserved.  SPDX-License-Identifier:
+-- (Apache-2.0 OR BSD-3-Clause)
 
--- CI script, invoked by both Travis and Appveyor
+-- CI script, compatible with all of Travis, Appveyor and Travis.
```
cocreature: That’s a lot of Travis :wink:

azure-pipelines.yml
```diff
@@ -38,4 +38,5 @@ steps:
       curl -sSL https://raw.github.com/ndmitchell/hlint/master/misc/travis.sh | sh -s examples/mini-compile/src
       curl -sSL https://get.haskellstack.org/ | sh
       stack setup > /dev/null
+      stack --no-terminal build
```
cocreature: Would be nice to add a comment somewhere for why we need this despite calling the same command in `CI.hs` (once you’ve finished debugging).

shayne-fletcher-da:
Now we are getting somewhere. Dialed the travis back and got us current with "Fix happy deps for Stack" 6/27.

azure-pipelines.yml
shayne-fletcher-da: The next commit knocks that line out.

shayne-fletcher-da: For the record, that particular `stack exec` was working because the `stack build` that preceded it caused `extra` to be installed. That is, was serendipitous rather than good design.

shayne-fletcher-da:
This PR brings `ghc-lib` level with `ghc` HEAD as of 7/3/2019.

hurryabit:
Very nice! Thanks.

PR #91, opened 2019-07-07
Upgrade to a more rigorous command-line parser

shayne-fletcher-da:
`ghc-lib-gen` gets a command-line parser via `optparse-applicative` and becomes a multi-file program.


CI.hs
```diff
@@ -71,8 +70,21 @@ main = do
...
+    --
+    -- ghc-lib-gen/src/Main.hs:9:1: error:
+    --     Could not find module ‘Ghclibgen’
+    --     Use -v to see a list of the files searched for.
+    --   |
+    -- 9 | import Ghclibgen
+    --   | ^^^^^^^^^^^^^^^^
+    --
+    -- Makes no sense to me.
+    --cmd "stack exec --no-terminal -- ghc -package=ghc-lib-parser -e \"print 1\""
```
shayne-fletcher-da: @cocreature Any insight you can offer into what's going wrong here would be appreciated!

cocreature: Looks like you solved this?

ghc-lib-gen.cabal
```diff
@@ -12,16 +13,36 @@ copyright:           Digital Asset 2018-2019
...
+                     , process >=1.6
+                     , filepath >=1.4
+                     , containers
+                     , directory
+                     , optparse-applicative
+                     , extra >=1.6
+  hs-source-dirs:    ghc-lib-gen/src
+  exposed-modules:   Ghclibgen
+                     GhclibgenOpts
+                     Paths_ghc_lib_gen
```
cocreature: I wouldn’t expose this, you aren’t actually using it outside of the lib.

ghc-lib-gen.cabal
```diff
@@ -12,16 +13,36 @@ copyright:           Digital Asset 2018-2019
...
                      , directory
+                     , optparse-applicative
                      , extra >=1.6
-  hs-source-dirs:      ghc-lib-gen/src
-  default-language:    Haskell2010
-  ghc-options: -Wmissing-signatures
+                     , ghc-lib-gen
+  other-modules:     Ghclibgen
+                     GhclibgenOpts
+                     Paths_ghc_lib_gen
```
cocreature: Those modules shouldn’t be listed in `other-modules` as you are getting them from the library. You probably have to remove the `hs-source-dirs` line and change `main-is` to `src/Main.hs` and/or use different `hs-source-dirs` for the executable and the library.

ghc-lib-gen.cabal
shayne-fletcher-da: You are seeing the detritus of my desperation trying to get Cabal happy and those failing tests restored :) Fixed.

ghc-lib-gen.cabal
shayne-fletcher-da: ^ Ditto but your advise on `ghc-lib-gen/src/Main.hs` appears to be the missing ingredient. Thanks!

PR #93, opened 2019-07-14
Simplify mini-hlint by using listify

shayne-fletcher-da:
Eliminate explicit tree traversal by using of `listify` from syb.


shayne-fletcher-da:
You are quite right. It's both a simplification and an  improvement.

PR #97, opened 2019-07-19
New example : strip locations; dump-parse tree

shayne-fletcher-da:
Example with a function to get output like `-ddump-parsed-ast` and use of uniplate to strip source locations from a parse tree. Also removes a lint from `ghc-lib-gen` options parsing.

### DIFF
--- a/examples/strip-locs/src/Main.hs
+++ b/examples/strip-locs/src/Main.hs
@@ -1,0 +1,123 @@
+-- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its
+-- affiliates. All rights reserved.  SPDX-License-Identifier:
+-- (Apache-2.0 OR BSD-3-Clause)
+
+{-# LANGUAGE PackageImports #-}
+{-# OPTIONS_GHC -Wno-missing-fields #-}
+
+module Main (main) where
+
+import "ghc-lib-parser" HsSyn
+import "ghc-lib-parser" Config
+import "ghc-lib-parser" DynFlags
+import "ghc-lib-parser" GHC.Platform
+import "ghc-lib-parser" StringBuffer
+import "ghc-lib-parser" Fingerprint
+import "ghc-lib-parser" Lexer
+import "ghc-lib-parser" RdrName
+import "ghc-lib-parser" ErrUtils
+import qualified "ghc-lib-parser" Parser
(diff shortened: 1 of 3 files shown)


ghc-lib-gen/src/GhclibgenOpts.hs
```diff
@@ -48,7 +49,8 @@ ghclibgenVersion =
 -- | A parser of a ghc-lib-gen target: `target := | "--ghc-lib-parser"
 -- | "--ghc-lib" | /* nothing */`.
 ghclibgenTarget :: Parser GhclibgenTarget
-ghclibgenTarget = fmap (fromMaybe Ghclib) $ optional (ghclibParser <|> ghclib)
+ghclibgenTarget =
+  fromMaybe Ghclib Control.Applicative.<$> optional (ghclibParser <|> ghclib)
```
hurryabit: Why did we fully qualify `<$>` here? It should already be in `Prelude`, isn't it?

examples/strip-locs/src/Main.hs
```diff
@@ -0,0 +1,123 @@
...
+    catchErrors act = handleGhcException reportErr
+                        (handleSourceError reportErr act)
+    reportErr e = do putStrLn $ "error : " ++ show e; return Nothing
+
+dumpParseTree :: DynFlags -> Located (HsModule GhcPs) -> IO ()
+dumpParseTree flags m =
+  dumpSDoc flags alwaysQualify Opt_D_dump_parsed_ast "" $ showAstData NoBlankSrcSpan m
+
+stripLocs :: Located (HsModule GhcPs) -> Located (HsModule GhcPs)
+stripLocs = transformBi $ const noSrcSpan
```
hurryabit: How elegant. ;)

ghc-lib-gen/src/GhclibgenOpts.hs
shayne-fletcher-da: Yes. I just mechanically applied the suggestion hlint provided. I meant to come back to that in a later commit.

examples/strip-locs/src/Main.hs
shayne-fletcher-da: There's no denying it, that's the coolest line of code in the repo :)

PR #103, opened 2019-07-26
Delete old artifacts starting CI and keep .tar.gz files

shayne-fletcher-da:
This PR modifies the CI script:
  - On starting up, old artifacts are deleted;
  - We don't delete the `.tar.gz` files created during the CI run.

These two things are convenient for running local builds.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -5,17 +5,26 @@
 -- CI script, compatible with all of Travis, Appveyor and Azure.
 
 import Control.Monad
+import Control.Exception
 import System.Directory
 import System.FilePath
+import System.IO.Error
 import System.IO.Extra
 import System.Info.Extra
 import System.Process.Extra
 import System.Time.Extra
+import Data.List
 
 main :: IO ()
 main = do
     when isWindows $
         cmd "stack exec -- pacman -S autoconf automake-wrapper make patch python tar --noconfirm"
+
+    -- Clear up any detritus left over from previous runs (useful when
+    -- building locally).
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -86,15 +93,21 @@ main = do
...
+      sDistCreateExtract target = do
         cmd "stack sdist ghc --tar-dir=."
-        [tarball] <- filter (isExtensionOf ".tar.gz") <$> getDirectoryContents "."
-        cmd $ "tar -xf " ++ tarball
-        return tarball
+        cmd $ "tar -xvf " ++ target ++ ".tar.gz"
+
+      rmFiles :: [FilePath] -> IO ()
+      rmFiles fs = forM_ fs $
+        \f -> removeFile f `catchIOError` (const  $ return ())
```
rohanjr: Can you check if the file exists and then remove it instead of catching the IO exception?

CI.hs
```diff
@@ -86,15 +93,21 @@ main = do
...
+
+      rmDirs :: [FilePath] -> IO ()
+      rmDirs ds = forM_ ds $
+        \d -> removeDirectoryRecursive d `catchIOError` (const $ return ())
```
rohanjr: Same here

CI.hs
shayne-fletcher-da: It's not worth it. The intent is just try and remove it and ignore any failure.

CI.hs
shayne-fletcher-da: As above.

PR #106, opened 2019-07-30
Tiniest of tweaks. Log actions when removing paths.

shayne-fletcher-da:
See paths getting removed on CI script running.
```
# Removing directory ghc
# Removing directory ghc-lib
# Removing directory ghc-lib-parser
# Removing file ghc-lib-0.20190730.tar.gz
# Removing file ghc-lib-parser-0.20190730.tar.gz
```

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -124,10 +124,7 @@ main = do
           =<< readFile' file
 
       removePath :: FilePath -> IO ()
-      removePath p = do
-        whenM (doesFileExist p) $ do
-          putStrLn ("# Removing file " ++ p)
-          removePathForcibly p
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -122,3 +122,12 @@ main = do
         writeFile file .
           replace "ghc-lib-parser" ("ghc-lib-parser == " ++ version)
           =<< readFile' file
+
+      removePath :: FilePath -> IO ()
+      removePath p = do
+        whenM (doesFileExist p) $ do
```
cocreature: Does the distinction between files and directories matter all that much? I would just use `doesPathExist` and only have a single case.

PR #115, opened 2019-08-27
Ci script upload

garyverhaegen-da:
This PR adds a flag to the CI.hs script to take care of uploading the tarballs to Bintray.

A few notes:
- I have renamed the existing `main` function to minimize the noise in the diff.
- I've made the script executable to avoid having to type out the (now quite long) stack invocation.
- I've duplicated the `cmd` function; there's probably a better way to handle this, but I did need a slightly different behaviour. Suggestions welcome.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -1,3 +1,4 @@
+#!/usr/bin/env stack runhaskell --package extra --package optparse-applicative --
 -- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its
 -- affiliates. All rights reserved.  SPDX-License-Identifier:
 -- (Apache-2.0 OR BSD-3-Clause)
--- a/CI.hs
+++ b/CI.hs
@@ -1,4 +1,4 @@
-#!/usr/bin/env stack runhaskell --package extra --package optparse-applicative --
+#!/usr/bin/env stack --resolver lts-12.10 runhaskell --package extra --package optparse-applicative --
 -- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its
 -- affiliates. All rights reserved.  SPDX-License-Identifier:
 -- (Apache-2.0 OR BSD-3-Clause)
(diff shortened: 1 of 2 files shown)


CI.hs
```diff
@@ -14,9 +15,62 @@ import System.Time.Extra
 import Data.List.Extra
 import Data.Time.Clock
 import Data.Time.Calendar
+import Data.Semigroup ((<>))
```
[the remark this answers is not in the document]
garyverhaegen-da: This one is actually not for the strings, but for the options on lines 26-28 (`Opts.progDesc` & `Opts.header`). I'm not sure what type they actually are but I suspect they're not mere `String`s.

CI.hs
```diff
@@ -14,9 +15,62 @@ import System.Time.Extra
...
+data Options = Options
+    { upload :: Bool }
+
+parseOptions :: Opts.Parser Options
+parseOptions = Options
+    <$> Opts.switch
+        ( Opts.long "upload-to-bintray"
+       <> Opts.help "If specified, will try uploading the sdists to Bintray, using credentials in BINTRAY_BASIC_AUTH env var.")
+
+doUpload :: String -> IO ()
```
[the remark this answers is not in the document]
garyverhaegen-da: Done.

CI.hs
```diff
@@ -14,9 +15,62 @@ import System.Time.Extra
...
 main = do
+    let opts = Opts.info (parseOptions Opts.<**> Opts.helper)
+                 ( Opts.fullDesc
+                <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
+                <> Opts.header "CI - CI script for ghc-lib" )
+    Options { upload = upload } <- Opts.execParser opts
+    version <- buildDists
+    when upload $ doUpload version
+
+data Options = Options
```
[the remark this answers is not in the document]
garyverhaegen-da: I'm not comfortable with the security implications of passing credentials as cli args. Env vars are usually what CI environments provide, which is why I went with this.

garyverhaegen-da:
Implemented the changes suggested by hlint — in particular changing `Options` to a `newtype`. Not convinced it matters here, but I don't see any reason to fight it either.

CI.hs
```diff
@@ -14,9 +15,64 @@ import System.Time.Extra
...
 main = do
+    let opts = Opts.info (parseOptions Opts.<**> Opts.helper)
+                 ( Opts.fullDesc
+                <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
+                <> Opts.header "CI - CI script for ghc-lib" )
+    Options upload  <- Opts.execParser opts
+    version <- buildDists
+    when upload $ bintrayUpload version
+
+newtype Options = Options Bool
```
cocreature:
```suggestion
newtype Options = Options { shouldUpload :: Bool }
```

Otherwise it is very unclear what this is supposed to represent.

CI.hs
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env stack runhaskell --package extra --package optparse-applicative --
```
cocreature: You might want to specify a resolver here to avoid this breaking when versions change. It probably picks up the resolver from the `stack.yaml` but I’m not entirely sure I trust that especially if you invoke it from a different working directory.

CI.hs
```diff
@@ -14,9 +15,64 @@ import System.Time.Extra
...
+        cmd $ concat ["curl -X POST",
+                          " -u", creds,
+                          " https://api.bintray.com/content/digitalassetsdk/ghc-lib/da-ghc-lib/", version, "/publish"]
+        where
+          cmd :: String -> IO ()
+          cmd x = do
+            let c = replace creds "***:***" x
+            putStrLn $ "\n\n# Running: " ++ c
+            hFlush stdout
+            (t, _) <- duration $ system_ x
```
cocreature: I would generally go for `callCommand` instead of `system_` if you don’t actually need to run it in a shell (afaict you don’t need to here).

CI.hs
garyverhaegen-da: Done

CI.hs
garyverhaegen-da: Added `--resolver lts-12.10` to match the stack.yaml file.

CI.hs
garyverhaegen-da: Done

PR #116, opened 2019-08-28
Remove a hs-source-dir from ghc-lib; add extra-libraries ffi

ghost:
Two "fixes" here:

(1) `ghcLibHsSrcDirs`: don't have `ghc-lib/stage0/libraries/ghci/build` there;
(2) Add `extra-libraries: ffi` to the library stanza's of both `ghc-lib-parser` and `ghc-lib`.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -84,8 +84,8 @@ ghcLibHsSrcDirs lib =
   let all = Set.fromList $
         [ "ghc-lib/stage1/compiler/build"
         -- , "ghc-lib/stage0/libraries/ghci/build"
-        -- No, don't include this. Have Cabal generate 'InfoTable.hs'
-        -- from the libraries/ghci, .hsc file. The hazel/bazel build
+        -- No, don't include this. Generate 'InfoTable.hs'
+        -- via the libraries/ghci, .hsc file. The hazel/bazel build
         -- breaks when we allow this source dir.
         ]
         ++ map takeDirectory cabalFileLibraries


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -439,7 +442,10 @@ generateGhcLibCabal = do
         ["    exposed-modules:"
         ,"        Paths_ghc_lib"
         ] ++
-        indent2 (nubSort nonParserModules)
+        indent2 (nubSort nonParserModules) ++
+        ["    extra-libraries:"
+        ,"        ffi"
```
cocreature: I thought that didn’t work properly when you are building with cabal/stack?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -83,7 +83,10 @@ ghcLibHsSrcDirs :: [Cabal] -> [FilePath]
 ghcLibHsSrcDirs lib =
   let all = Set.fromList $
         [ "ghc-lib/stage1/compiler/build"
-        , "ghc-lib/stage0/libraries/ghci/build"
+        -- , "ghc-lib/stage0/libraries/ghci/build"
```
cocreature: I assume cabal still works fine if we omit this?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -83,7 +83,10 @@ ghcLibHsSrcDirs :: [Cabal] -> [FilePath]
...
+        -- No, don't include this. Have Cabal generate 'InfoTable.hs'
```
cocreature: FWIW, those comments are somewhat inconsistent. In the Bazel case, Cabal does not come into play.

ghc-lib-gen/src/Ghclibgen.hs
cocreature: Too many branches :slightly_smiling_face:

PR #117, opened 2019-08-28
More postgen refinements

ghost:
Remove some intermediate generated files to avoid ambiguities bazel build.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -378,9 +378,9 @@ commonBuildDepends =
 libBinParserModules :: IO ([Cabal], [Cabal], [String])
 libBinParserModules = do
     lib <- mapM readCabalFile cabalFileLibraries
-    bin <- (:[]) <$> readCabalFile cabalFileBinary
+    bin <- readCabalFile cabalFileBinary
     parserModules <- calcParserModules
-    return (lib, bin, parserModules)
+    return (lib, [bin], parserModules)
 
 -- | Call this after cabal file generation. These files are generated
 -- from '.hsc' files in the source tree and we prefer to ship those in


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -374,6 +374,29 @@ commonBuildDepends =
   , "hpc == 0.6.*"
   ]
 
+-- | This utility factored out to avoid repetion.
+libBinParserModules :: IO ([Cabal], [Cabal], [String])
+libBinParserModules = do
+    lib <- mapM readCabalFile cabalFileLibraries
+    bin <- (:[]) <$> readCabalFile cabalFileBinary
```
cocreature:
```suggestion
    bin <- readCabalFile cabalFileBinary
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -374,6 +374,29 @@ commonBuildDepends =
...
+    parserModules <- calcParserModules
+    return (lib, bin, parserModules)
```
cocreature:
```suggestion
    return (lib, [bin], parserModules)
```
Seems a bit easier to read for me this way.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -374,6 +374,29 @@ commonBuildDepends =
...
+
+-- | Call this after cabal file generation. These files are generated
+-- from '.hsc' files in the source tree and we prefer to ship those in
+-- the sdists rather than these. If we don't remove them, some
+-- ambiguity results. Cabal and stack are OK with that but bazel gets
```
cocreature: Note that even for Cabal and Stack it makes sense to not ship them. hsc2hs is platform dependent no including the generated files isn’t going to work in general.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -374,6 +374,29 @@ commonBuildDepends =
...
+-- messed up.
+removeGeneratedIntermediateFiles :: IO ()
+removeGeneratedIntermediateFiles = do
```
cocreature: How hard is it to automate this list? Can we just glob for `.hsc` files and remove all corresponding `.hs` files?

cocreature: Absolutely, I’m also fine with just ignoring this and not even raising an issue :slightly_smiling_face:

PR #122, opened 2019-08-29
Enable Azure pipelines on the da-ghc-8.8.1 branch

cocreature:
Given that this is the branch we care about for our own code it seems useful to enable CI here.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,6 +8,7 @@ trigger:
   branches:
     include:
     - master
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
Note that this will run CI on changes to this branch, but will not run CI on PRs targeting this branch. Is that the intended behaviour?

cocreature:
Good catch, thanks! That was definitely not indended and I changed it now.

PR #130, opened 2019-09-05
Now we've actually done it, we add tested-with GHC==8.8.1

shayne-fletcher-da:
This affects the contents of generated `ghc-lib-parser.cabal` and `ghc-lib.cabal` such that their `tested-with` fields contain 8.8.1.


shayne-fletcher-da:
> 🎉 You might also want to change `8.6.3` to `8.6.5`

Will do!

PR #131, opened 2019-09-05
Allow testing with different compiler versions

shayne-fletcher-da:
This PR adds an optional `--resolver` flag to `CI.hs` allowing the compiler used to build and test`ghc-lib` packages to override the compiler determined by `stack.yaml`.

Using this newly added capability, extend the Azure strategy matrix  so that  we can now test all configurations against both `8.4.4` and `8.6.5` (and when there's a resolver for `8.8.1` we can add that too).

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -23,18 +23,26 @@ import qualified System.Exit as Exit
 
 main :: IO ()
 main = do
-    let opts = Opts.info (parseOptions Opts.<**> Opts.helper)
-                 ( Opts.fullDesc
-                <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
-                <> Opts.header "CI - CI script for ghc-lib" )
-    Options { upload, ghcFlavor }  <- Opts.execParser opts
-    version <- buildDists ghcFlavor
+    let opts =
+          Opts.info (parseOptions Opts.<**> Opts.helper)
+          ( Opts.fullDesc
+            <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
+            <> Opts.header "CI - CI script for ghc-lib"
+          )
+    options@Options { upload, ghcFlavor, resolver }  <- Opts.execParser opts
+    version <- buildDists ghcFlavor resolver
     when upload $ bintrayUpload version
 
 data Options = Options
     { upload :: Bool
     , ghcFlavor :: GhcFlavor
-    }
+    , resolver :: Maybe String -- If 'Just _', override stack.yaml.
+    } deriving (Show)
+
+stackResolverOpt :: Maybe String -> String
+stackResolverOpt = \case
+  Just resolver -> "--resolver " ++ resolver
+  Nothing -> ""
 
 data GhcFlavor = Ghc881 | DaGhc881 | GhcMaster
   deriving (Eq, Show)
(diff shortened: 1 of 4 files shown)


CI.hs
```diff
@@ -62,30 +74,33 @@ parseOptions = Options
...
       Nothing ->
-        Exit.die $ unlines ["Error: Cannot upload without BINTRAY_BASIC_AUTH.",
-                            "To set the environment variable, run:",
-                            "    BINTRAY_BASIC_AUTH=<creds> ./CI.hs --upload-to-bintray",
-                            "where <creds> should be of the form:",
-                            "  fname.lname@digitalassetsdk:0123456789abcdef0123456789abcdef01234567",
-                            "You can find your API key (the part after the colon) at:",
-                            "  https://bintray.com/profile/edit",
-                            "after logging in. (The username is also displayed on that page.)"]
+        Exit.die $ unlines [
```
cocreature: It would be nice if you could separate reformatting from actual changes. Mixing the two in the same PR makes it a bit hard to see what actually changed in a PR.

CI.hs
```diff
@@ -96,11 +111,14 @@ bintrayUpload version = do
...
-buildDists ghcFlavor = do
+buildDists :: GhcFlavor -> Maybe String -> IO String
+buildDists ghcFlavor resolver = do
+    -- One of "" (use 'stack.yaml') or, "--resolver=xxx".
+    let resolverFlag=stackResolverOpt resolver
+
     -- Get packages missing on Windows needed by hadrian.
     when isWindows $
-        cmd "stack exec -- pacman -S autoconf automake-wrapper make patch python tar --noconfirm"
+        cmd $ "stack exec "  ++ resolverFlag ++ " -- pacman -S autoconf automake-wrapper make patch python tar --noconfirm"
```
cocreature: Maybe it would make sense to define something like `let stackCmd cmd = cmd $ "stack " ++ resolverFlag ++ " " ++ cmd` at the beginning of this function and then use that everywhere. That makes it a bit easier to set default flags that should be passed to all Stack invocations.

azure-pipelines.yml
```diff
@@ -18,35 +18,95 @@ pr:
...
       ghc-flavor: "da-ghc-8.8.1"
+      resolver: "lts-14.3"
+
+    # ghc-lib master (vanilla)
+
+    # ghc-8.4.4
+    linux-ghc-master-844:
+      image: "Ubuntu 16.04"
+      ghc-flavor: "ghc-master"
+      resolver: "lts-12.26"
```
cocreature: I am not entirely sold on just specifying the resolver here. That breaks quickly once you need some kind of extra-deps. Maybe we should just add a `stack-8.4.4.yaml` file to this repo and then instead of passing `--resolver` we can pass `--stack-yaml` which gives us more flexibility?

CI.hs
shayne-fletcher-da: Yeah. Sorry! 🙇

CI.hs
shayne-fletcher-da: Indeed. I'll look into cleaning this up today.

azure-pipelines.yml
shayne-fletcher-da: I hear you but my understanding is, if a resolver is given it *overrides* the resolver in 'stack.yaml' and what's in the 'stack.yaml' otherwise applies including 'extra-deps' so I'm not sure that there is much extra flexibility to be had by doing that. On the other hand, if it proves useful in the future, the code is now easily adaptable to this approach so how about we take a wait and see?

cocreature: Not quite sure on the behavior of `resolver` but the extra flexibility becomes useful if you actually need different extra-deps per resolver which can happen quite easily.

shayne-fletcher-da: Ok. That's the bit that I wasn't sure in fact came up (extra deps on a per config basis). That being the case, agreed. Let's be lazy though and do it down the road when we have to :)

PR #135, opened 2019-09-14
Sync with upstream

shayne-fletcher-da:
Sync to `bd079345f633e01e8d4ff76aef39cb2aa50af6a9`.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -44,7 +44,8 @@ data GhcFlavor = Ghc881 | DaGhc881 | GhcMaster String
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current =    "270fbe8512f04b6107755fa22bdec62205c0a567" -- 09/09/2019
+current =    "bd079345f633e01e8d4ff76aef39cb2aa50af6a9" -- 09/14/2019
+          -- "270fbe8512f04b6107755fa22bdec62205c0a567" -- 09/09/2019
           -- "b55ee979d32df938eee9c4c02c189f8be267e8a1" -- 09/06/2019
           -- "11679e5bec1994775072e8e60f24b4ce104af0a7" -- 09/03/2019
 
@@ -154,7 +155,6 @@ buildDists ghcFlavor resolver = do
             cmd "cd ghc && git -c \"user.name=Cookie Monster\" -c \"user.email=cookie.monster@seasame-street.com\" merge --no-edit upstream/da-master-8.8.1 upstream/da-unit-ids-8.8.1"
         GhcMaster hash -> cmd $ "cd ghc && git checkout " ++ hash
     cmd "cd ghc && git submodule update --init --recursive"
-    appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
 
     -- Feedback on the compiler used for ghc-lib-gen.
     stack "exec -- ghc --version"
@@ -172,6 +172,11 @@ buildDists ghcFlavor resolver = do
         pkg_ghclib_parser = "ghc-lib-parser-" ++ version
 
     -- Make and extract an sdist of ghc-lib-parser.
+    appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
+    -- Temporary (09/14/2012) : Until
+    -- https://gitlab.haskell.org/ghc/ghc/merge_requests/1716 lands.
+    appendFile "ghc/hadrian/stack.yaml" $ unlines ["allow-newer: True"]
+    -- Feedback on the compiler used for ghc-lib-gen.
     stack $ "exec -- ghc-lib-gen ghc --ghc-lib-parser " ++ ghcFlavorOpt ghcFlavor
     patchVersion version "ghc/ghc-lib-parser.cabal"
     mkTarball pkg_ghclib_parser resolverFlag
@@ -182,6 +187,10 @@ buildDists ghcFlavor resolver = do
     -- Make and extract an sdist of ghc-lib.
     cmd "cd ghc && git checkout ."
     appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
+    -- Temporary (09/14/2012) : Until
+    -- https://gitlab.haskell.org/ghc/ghc/merge_requests/1716 lands.
+    appendFile "ghc/hadrian/stack.yaml" $ unlines ["allow-newer: True"]
+    -- Feedback on the compiler used for ghc-lib-gen.
     stack $ "exec -- ghc-lib-gen ghc --ghc-lib " ++ ghcFlavorOpt ghcFlavor
     patchVersion version "ghc/ghc-lib.cabal"
     patchConstraint version "ghc/ghc-lib.cabal"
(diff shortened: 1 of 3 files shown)


CI.hs
```diff
@@ -172,6 +172,11 @@ buildDists ghcFlavor resolver = do
         pkg_ghclib_parser = "ghc-lib-parser-" ++ version
 
     -- Make and extract an sdist of ghc-lib-parser.
+    appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
+    -- Temporary (09/14/2012) : Until
```
cocreature: I don’t think that’s a good idea. The MR is already approved and assigned to marge. I don’t see what we gain by patching this now only to have to undo the patch in a few days.

shayne-fletcher-da: As soon as it lands I'll undo it. For now though I can't uncover the other issues without it :)

shayne-fletcher-da: ^ I promise!

cocreature: My point is why don’t we just wait until it’s patched upstream? We don’t even use the GHC master version in DAML so waiting a few days seems totally reasonable.

shayne-fletcher-da: Because it's my experience that keep it up-to-date incrementally is easier than letting all the breaking changes stack up! I mean, it's no big deal but keep it working helps with my anxiety.

shayne-fletcher-da: In this case, I could remove this patch and just commit the rest but of course, it would then show as broken in CI. I don't know if that's preferable?

cocreature: Fair enough but I’m not entirely convinced that we really need to keep the frequency we have right now.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -166,33 +167,33 @@ extraFiles ghcFlavor =
...
+cHeaders :: GhcFlavor -> [String]
+cHeaders ghcFlavor =
   [ "includes/ghcconfig.h"
   , "includes/MachDeps.h"
   , "includes/CodeGen.Platform.hs"
-  , "compiler/nativeGen/*.h"
-  , "compiler/utils/*.h"
-  , "compiler/*.h"
-  ]
+  , "compiler/HsVersions.h"
```
cocreature: Are `HsVersions` and `Unique` the only headers in `compiler` for GHC 8.8 or why are we switching from a wildcard for 8.8 to listing the two explicitly?

CI.hs
shayne-fletcher-da: The frequency with which I sync with upstream? I tend to do so once a week. This one cost me ~3 or 4 hours. I've tried many strategies - this one is working best for me 🤷‍♀

shayne-fletcher-da:
Something quite odd going on here on Windows :-/

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -447,9 +451,12 @@ generateGhcLibCabal ghcFlavor = do
         ,"data-files:"] ++
         indent dataFiles ++
         ["extra-source-files:"] ++
-        -- Remove Config.hs, Version.hs, Parser.hs and Lexer.hs from
-        -- the list of extra source files here.
-        indent (reverse (drop 4 $ reverse $ extraFiles ghcFlavor)) ++ indent cHeaders ++
+         ( let numFiles = if ghcFlavor == GhcMaster then 4 else 3 in
+             -- Remove Config.hs, Version.hs, Parser.hs and Lexer.hs from
+             -- the list of extra source files here.
+             indent (reverse (drop numFiles $ reverse $ extraFiles ghcFlavor)) ++ indent (cHeaders ghcFlavor)
```
cocreature: This code is trying to be clever and reuse `extraFiles` but I think it actually ends up being harder to read and more fragile than just listing the files that you want to remove here or alternatively factor it into a separate list `generatedFiles ghcFlavor` and turn `extraFiles` into something like `extraFiles ghcFlavor = … ++ generatedFiles ghcFlavor`. I also don’t understand why we remove those files for `ghc-lib-parser.cabal` but keep them for `ghc-lib.cabal`.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,4 +586,5 @@ generatePrerequisites ghcFlavor = do
   -- of the way.
   removeFile "compiler/parser/Lexer.x"
```
cocreature: If we factor the generated files into a separate list as I suggested above we could also reuse it here instead of replicating the logic of which files are generated for which GHC version.

shayne-fletcher-da:
You know what we can do? Since the work has been done, we can (1) keep this PR open, (2) wait for the MR to land and when it has (3) update the PR and then land it? How does that sit with you?

shayne-fletcher-da:
> LGTM, I’m happy to merge this with the patch included.

Awesome. Thank-you.

> I was just wondering if we might be making these GHC updates more painful (or at least time consuming) than they already are by a) making them fairly frequent

No, the opposite in fact.

> and b) working around issues that will be fixed upstream fairly soon only to then have to undo them.

Well, experience again dictates that once a bug is identified, there can be a long period of latency before it actually gets resolved (witness the weeks around Zurihac).

> But given that so far you’re doing all the GHC bumps, I’m more than happy to leave this up to you.

Like Peter Venkman (Ghostbusters) said "Trust me. I'm a scientist" 😉

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher-da: I looked into it. Yes, they are the only headers in 8.8. I'm implementing a general policy of removing wild-cards (following on from what we learned with our rts related include work).

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher-da: Isn't it the case that they go into `ghc-lib-parser.cabal` but NOT `ghc-lib.cabal`? Checking... yes.

I take your point about reusing `extraFiles`. It's an artifact of evolution. It certainly can be improved.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher-da: Yes. Noted. As stated above, I'll raise an issue to clean this up.

shayne-fletcher-da:
Windows is failing due to a known issue with naming a file `Con.hs`. Issue is here https://gitlab.haskell.org/ghc/ghc/issues/17187 and fix is in the merge queue https://gitlab.haskell.org/ghc/ghc/merge_requests/1714

shayne-fletcher-da:
@cocreature , issue for the abuse of extra-files raised https://github.com/digital-asset/ghc-lib/issues/136.

shayne-fletcher-da:
So given there are two MRs in the queue (https://gitlab.haskell.org/ghc/ghc/merge_requests/1716 and https://gitlab.haskell.org/ghc/ghc/merge_requests/1714) have removed the manual work-around for the first and am putting this PR in hiatus for now as suggested by @cocreature . Will update when the above is landed.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher-da: Fix coming.

cocreature:
Both PRs have been merged upstream now afaict.

PR #137, opened 2019-09-16
Sync to 7915afc6bb9539a4534db99aeb6616a6d145918a

shayne-fletcher-da:
We need to consider this before landing it. The long and the short of it is that the referenced commit has drastically changed modularity in the GHC API such that to run the parser now requires 534 files whereas before that commit, it only needed 200 or so.


shayne-fletcher-da:
I've been in touch with the author (Sebastian Graf). He's going to look into refactoring his commit.

shayne-fletcher-da:
Expecting changes that are going to render much of this moot. Closing out this PR without merging but will keep the branch in case I need to refer back to it.

PR #140, opened 2019-09-23
Don't reference stage1 files, strictly stage0 only

shayne-fletcher-da:
Up until now, include and source directories have included a mix of stage0 and stage1 artifacts. I'm a little sad and embarrassed to report there's been no good (rhyme or) reason for this - all I can say about it is it's an artifact of the evolution of this program over time "hacking" 'til it works 🗑️🔥 

This PR finally redresses this issue. Now we strictly access stage0 artifacts and never, never ask for a stage1 artifact. At the least an unnecessary source of complexity is removed. It may be the case that generation of the cabal files is made a little faster by this development (but I can't say I've noticed a big difference).

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -46,7 +46,6 @@ ghcLibParserIncludeDirs ghcFlavor =
   [ "includes" -- ghcconfig.h, MachDeps.h, CodeGen.Platform.hs
   , "ghc-lib/generated"
   , "ghc-lib/stage0/compiler/build"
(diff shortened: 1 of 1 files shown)


shayne-fletcher-da:
Yes. I believe stage0 corresponds to the Make scheme's stage1.

cocreature:
It sounds like the Hadrian model is “everything built by stage X is in stageX” as opposed to “everything that makes up stage X is in stage X”. So since the stage 1 compiler is built by stage 0 it is in stage0.

shayne-fletcher-da:
That's exactly it. Well put.

Issue #149, opened 2019-10-09
Check our GHC_STAGE handling


shayne-fletcher-da:
This ghc [commit](https://gitlab.haskell.org/ghc/ghc/commit/817c1a947d50a070613815e02a530f83d9078eb8) does a bit more than rename the manifest constant `STAGE` to `GHC_STAGE` - it looks like it sets the constant in `ghc_platform.h` rather than from the command line.

We have two things to consider in ghc-lib-gen : `applyPatchStage` (probably fine) and whether we should be passing `-DGHC_STAGE` to the `cpp-options` in the ghc-lib-parser and ghc-lib cabal files. I suspect not because I'm seeing lots of warnings like this:
```
ghc-lib> /Users/shaynefletcher/tmp/ghc-lib/ghc-lib/ghc-lib/stage0/lib/ghcplatform.h:4:9: error:
ghc-lib>      warning: 'GHC_STAGE' macro redefined [-Wmacro-redefined]
ghc-lib>   |
ghc-lib> 4 | #define GHC_STAGE 1
ghc-lib>   |         ^
ghc-lib> #define GHC_STAGE 1
```

cocreature:
> This ghc [commit](https://gitlab.haskell.org/ghc/ghc/commit/817c1a947d50a070613815e02a530f83d9078eb8) does a bit more than rename the manifest constant `STAGE` to `GHC_STAGE`

It doesn’t seem to do that at all, that’s part of a separate commit https://gitlab.haskell.org/ghc/ghc/commit/0c0a15a8541a9119536ee17849d8390cb21ab764.

As for the CPP define, we definitely shouldn’t set that anymore. We either need to check that the way it’s defined in `ghcplatform.h` is fine with all `#if` guards (we patch out some that were problematic in the past but there are probably others) in `ghc-lib`, patch `ghcplatform.h` to set it to something different if that fixes the `#if` guards or patch out all of the remaining guards to whatever we want them to be.

shayne-fletcher-da:
On master, `ghcplatform.h` sets `GHC_STAGE = 1` and we no longer are required to pass pass `-DGHC_STAGE=2` to `cpp-options` to get a successful build. All usages of `getOrSetLibHSghc*` require `GHC_STAGE >= 2`. So, it looks like we are no longer required to `applyPatchStage` on master and can drop the `-DGHC_STAGE=2` from the cabal files.

PR #150, opened 2019-10-09
Sync upstream

shayne-fletcher-da:
Upstream sweep to `d584e3f08cfee6e28b70bf53c573d86e44f326f8`.

Issue raised on whether we should be continuing to set the (now called) `GHC_STAGE` constant in the `cpp-options` field of the generated cabals here : https://github.com/digital-asset/ghc-lib/issues/149.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -383,7 +383,13 @@ mangleCSymbols _ = do
 -- by using getOrSetLibHSghc for the FastString table.
 applyPatchStage :: GhcFlavor -> IO ()
 applyPatchStage ghcFlavor =
-    forM_ [ "compiler/ghci/Linker.hs"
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -552,6 +555,10 @@ ghciDef :: GhcFlavor -> String
 ghciDef GhcMaster = ""
 ghciDef _ = "-DGHCI"
 
+ghcStageDef :: GhcFlavor -> String
+ghcStageDef GhcMaster = "-DGHC_STAGE=2"
```
cocreature: This is definitely not the right thing to do now that this is defined in the header. We shouldn’t pass any define flags now, see #149 for some alternatives.

shayne-fletcher-da:
> If so, it would be good to test this on the haskell-ide-engine testsuite again and confirm that we haven’t created any new segfaults.
Still mean to try to test this if it's not crazy work to do so.

PR #153, opened 2019-10-22
Sync upstream

shayne-fletcher-da:
Sync to `2d2cc76ffb781d01c800608cd8be05cca67ac4c0`.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -43,7 +43,7 @@ cabalFileLibraries =
 -- | C-preprocessor "include dirs" for 'ghc-lib-parser'.
 ghcLibParserIncludeDirs :: GhcFlavor -> [FilePath]
 ghcLibParserIncludeDirs ghcFlavor =
-  [ "includes" ] ++ -- ghcconfig.h, MachDeps.h, CodeGen.Platform.hs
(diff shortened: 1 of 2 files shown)


shayne-fletcher-da:
Current set of failures all seem to be network related.

shayne-fletcher-da:
The change is good. CI is looking a bit broke - network issues affecting linux and Windows.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -208,6 +208,7 @@ cHeaders :: GhcFlavor -> [String]
 cHeaders ghcFlavor =
   [ "includes/ghcconfig.h"
   , "includes/MachDeps.h"
+  , "includes/stg/MachRegs.h"
```
cocreature: Shouldn’t this change depend on the STG version? I don’t see how this could have changed for older versions.

shayne-fletcher-da: Not an oversight. The inclusion of a new set of constants to support the LLVM backend highlighted a bug in our inclusion scheme. The contents of `includes/CodeGen.Platform.hs` includes ` stg/MachRegs.h`. The former is in the `extra-sources` of ghc-lib targets but not the latter which we've been satisfying by provision from the compiler. Now we include `MachRegs.h` (as we should) along with `MachDeps.h`.

cocreature: Sounds good :+1:

PR #160, opened 2019-12-04
Mark ghc-lib as exposed: False

### LINKED ISSUE #158: Hidden by default
The README states that `ghc-lib` and `ghc-lib-parser` are currently not hidden by default because this requires newer versions of Cabal. Can this decision be revisited, now that `cabal-install` 3.0.0.0 is widely used?

cocreature:
I haven’t looked into the actual Cabal version that added support for
this but it is documented since 2008 so it seems very reasonable to
rely on this
https://github.com/haskell/cabal/commit/214197f41840c1901e2ddb6dd37391f557964d82.

fixes #158


vaibhavsagar:
Thank you!

PR #161, opened 2019-12-21
Adapt master to work with older ghc-prim

shayne-fletcher-da:
Commit [`705a16df02411ec2445c9a254396a93cabe559ef`](https://gitlab.haskell.org/ghc/ghc/commit/705a16df02411ec2445c9a254396a93cabe559ef) on 12/3/2019 induces a reliance on a version of `ghc-prim` that is not available to us (we are stuck with the `ghc-prim` that ships with the building compiler).

In the absence of a better patch, this one at least enables us to continue producing `ghc-lib` builds against `HEAD`.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -4,6 +4,7 @@
 
 module Ghclibgen (
     applyPatchHeapClosures
+  , applyPatchGhcPrim
   , applyPatchDisableCompileTimeOptimizations
   , applyPatchRtsIncludePaths
   , applyPatchStage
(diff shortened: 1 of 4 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -325,6 +326,42 @@ applyPatchHeapClosures _ = do
...
+-- (https://gitlab.haskell.org/ghc/ghc/commit/705a16df02411ec2445c9a254396a93cabe559ef)
+applyPatchGhcPrim :: GhcFlavor -> IO ()
+applyPatchGhcPrim ghcFlavor = do
+    let tysPrim = "compiler/prelude/TysPrim.hs"
+    when (ghcFlavor == GhcMaster) (
+      writeFile tysPrim .
+          replace
+              "bcoPrimTyConName              = mkPrimTc (fsLit \"BCO\") bcoPrimTyConKey bcoPrimTyCon"
+              "bcoPrimTyConName              = mkPrimTc (fsLit \"BCO#\") bcoPrimTyConKey bcoPrimTyCon"
+        =<< readFile' tysPrim
```
aherrmann-da: Does the [`LiftedRep --> UnliftedRep`](https://gitlab.haskell.org/ghc/ghc/commit/705a16df02411ec2445c9a254396a93cabe559ef#0188aac58d5dfb76292827933bb504a964ed7c01_1058_1061) also need to be patched?

shayne-fletcher-da: Wow. Well spotted. Yes it does and I'm at a loss to explain how that got through without it! Fixing and testing it now. Thank-you!

PR #162, opened 2019-12-21
Add support for ghc-8.10.1

shayne-fletcher-da:
Support for ghc-8.10.1 (at this time pulls the ghc-8.8.1-alpha2 branch).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,9 +20,9 @@ strategy:
   # Limit number of executors used so other pipelines can run too
   maxParallel: 10
   matrix:
-    # ghc-lib v8.8.1 (vanilla)
+    # -- compiler : ghc-8.4.4
 
-    # ghc-8.4.4
+    # ---- ghc-flavor : ghc-8.8.1
     linux-ghc-881-844:
       image: "Ubuntu 16.04"
       ghc-flavor: "ghc-8.8.1"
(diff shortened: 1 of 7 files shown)


azure-pipelines.yml
```diff
@@ -103,10 +132,7 @@ strategy:
     #   image: "vs2017-win2016"
     #   ghc-flavor: "da-ghc-8.8.1"
     #   resolver: "nightly-2019-09-26"
-    mac-da-ghc-881-881:
-      image: "macOS-10.13"
-      ghc-flavor: "da-ghc-8.8.1"
-      resolver: "nightly-2019-09-26"
+    # ---- ghc-flavor : da-ghc-8.10.1 (doesn't exist)
```
aherrmann-da: Why is this one removed?

ghc-lib-gen/src/GhclibgenOpts.hs
```diff
@@ -76,7 +76,8 @@ ghcFlavorOpt = option readFlavor
 
 readFlavor :: ReadM GhcFlavor
 readFlavor = eitherReader $ \case
+    "ghc-8.10.1" -> Right Ghc8101
     "ghc-8.8.1" -> Right Ghc881
     "da-ghc-8.8.1" -> Right DaGhc881
     "ghc-master" -> Right GhcMaster
-    flavor -> Left $ "Failed to parse ghc flavor " <> show flavor <> " expected ghc-8.8.1 or da-ghc-8.8.1"
+    flavor -> Left $ "Failed to parse ghc flavor " <> show flavor <> " expected ghc-8.8.1, da-ghc-8.8.1 or ghc-8.10.1"
```
aherrmann-da: Why is `ghc-master` not listed as an expected input?

azure-pipelines.yml
shayne-fletcher-da: Argh. Just a mistake. Nice catch and thank-you. I don't know why but I struggle with that YAML format and keeping this matrix straight!

ghc-lib-gen/src/GhclibgenOpts.hs
shayne-fletcher-da: Good point. Just an oversight - I'll fix it. (actually, even then it's an approximation. You can pass a SHA in that position too)

Issue #142, opened 2019-09-28
ghci ghc-lib-parser load test fails on windows with ghc-8.8.1


shayne-fletcher-da:
All Windows flavors (`ghc-8.8.1`, `da-ghc-8.81`, `ghc-master`) compiled with `ghc-8.8.1` fail the load `ghc-lib-parser` in ghci test and are disabled.
```
2019-09-27T22:58:00.4061936Z # Running: stack --resolver nightly-2019-09-26 --no-terminal exec -- ghc -ignore-dot-ghci -package=ghc-lib-parser -e "print 1"
2019-09-27T22:58:00.4758885Z stty: 'standard input': Inappropriate ioctl for device
2019-09-27T22:58:00.7109822Z [0mStack has not been tested with GHC versions above 8.6, and using 8.8.1, this may fail[0m
2019-09-27T22:58:00.8196477Z [0mStack has not been tested with Cabal versions above 2.4, but version 3.0.0.0 was found, this may fail[0m
2019-09-27T22:58:01.6821466Z D:\a\1\s\.stack-work\install\a5e678e4\lib\x86_64-windows-ghc-8.8.1\ghc-lib-parser-0.20190927-FUwiPDktS2BEO3PvZtLNsC\HSghc-lib-parser-0.20190927-FUwiPDktS2BEO3PvZtLNsC.o: unhandled PEi386 relocation type 0
2019-09-27T22:58:01.6853587Z ghc.EXE: unable to load package `ghc-lib-parser-0.20190927'
```

mvoidex:
Same error with `haskell-src-exts-1.21.1` under windows, ghc-8.8.1

shayne-fletcher-da:
Perhaps related :
  - https://ci.appveyor.com/project/ndmitchell/hoogle#L5349
 - https://github.com/ghc/ghc/blob/70e56b272492b65e41a149ec39a939e794fea66b/rts/linker/PEi386.c#L1990
- https://gitlab.haskell.org/ghc/ghc/issues/9907?cversion=0&cnum_hist=5#note_104251

shayne-fletcher-da:
Raised issue with GHC HQ https://gitlab.haskell.org/ghc/ghc/issues/17599.

shayne-fletcher-da:
This issue is said to be an instance of https://gitlab.haskell.org/ghc/ghc/issues/17575 and it is expected that it will be resolved by 8.8.2.

flip111:
[...]

PR #168, opened 2020-01-05
Restrict skipping Windows ghci load tests to 8.8.1 exclusively

shayne-fletcher-da:
The discussion in https://gitlab.haskell.org/ghc/ghc/issues/17599 indicates the issue described in https://github.com/digital-asset/ghc-lib/issues/142 is expected to be resolved in ghc-8.8.2. Accordingly, restrict skipping the failing Windows ghci load tests to 8.8.1 exclusively.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -237,7 +237,7 @@ buildDists ghcFlavor resolver = do
     stack "--no-terminal exec -- strip-locs examples/mini-compile/test/MiniCompileTest.hs"
     stack "--no-terminal exec -- mini-compile examples/mini-compile/test/MiniCompileTest.hs"
 
-#if __GLASGOW_HASKELL__ >= 808 && \
+#if __GLASGOW_HASKELL__ == 808 && \
     __GLASGOW_HASKELL_PATCHLEVEL1__ == 1 && \
     defined (mingw32_HOST_OS)
     -- Skip these tests on ghc-8.8.1 (exclusively). See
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -237,8 +237,10 @@ buildDists ghcFlavor resolver = do
     stack "--no-terminal exec -- strip-locs examples/mini-compile/test/MiniCompileTest.hs"
     stack "--no-terminal exec -- mini-compile examples/mini-compile/test/MiniCompileTest.hs"
 
-#if __GLASGOW_HASKELL__ >= 808 && defined (mingw32_HOST_OS)
-    -- Skip these tests on ghc-8.8.1 (and greater) for now. See
+#if __GLASGOW_HASKELL__ >= 808 && \
```
hurryabit: That would skip the test for `8.10.1` as well. Is that what we want?

cocreature: It looks like this should probably be `if __GLASGOW_HASKELL__ < 808 || _GLASGOW_HASKELL_ == 808 && __GLASGOW_HASKELL_PATHLEVEL1__ == 1 …`

shayne-fletcher-da: Yes. That's an error. Thank-you.

shayne-fletcher-da:
@cocreature @hurryabit Re-review please.

CI.hs
```diff
@@ -237,8 +237,10 @@ buildDists ghcFlavor resolver = do
     stack "--no-terminal exec -- strip-locs examples/mini-compile/test/MiniCompileTest.hs"
     stack "--no-terminal exec -- mini-compile examples/mini-compile/test/MiniCompileTest.hs"
 
-#if __GLASGOW_HASKELL__ >= 808 && defined (mingw32_HOST_OS)
-    -- Skip these tests on ghc-8.8.1 (and greater) for now. See
+#if __GLASGOW_HASKELL__ == 808 && \
```
cocreature: This looks like you are missing GHC < 808 where this also doesn’t work.

PR #170, opened 2020-01-05
Disable build optimizations with --fast

shayne-fletcher-da:
Use `stack build TARGET --fast` to disable build optimizations instead of writing a stanza to `stack.yaml`.


shayne-fletcher-da:
That's a good point. I was trying to make life a little simpler but I think given this I'll abandon the change.

PR #173, opened 2020-01-11
Sync head and add more stack flags

### LINKED ISSUE #169: Add some verbosity args to CI.Hs
Extend the set of command line flags to CI.hs to include:
  - [x]  `--stack-verbose`:
       Set stack's `--verbosity` : One of "silent", "error", "warn", "info", debug"  e.g. `stack build --verbosity=debug`.
  - [x] `--cabal-verbose`
      Pass to `stack build`. Ask Cabal to be verbose  e.g. `stack build --cabal-verbose`
  - [x]  `--ghc-options`:
     Pass to `stack build` e.g. `--ghc-options="-v3"` (get ghc to be verbose).

shayne-fletcher-da:
- Sync to `923a127205dd60147453f4420614efd1be29f070`;
- Add `--verbosity=xxx`, `--cabal-verbose` and `--ghc-options=xxx` arguments for CI.hs ( fixes #169 ).

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -1,5 +1,5 @@
--- Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its
--- affiliates. All rights reserved.  SPDX-License-Identifier:
+-- Copyright (c) 2019-2020, Digital Asset (Switzerland) GmbH and/or
+-- its affiliates. All rights reserved.  SPDX-License-Identifier:
 -- (Apache-2.0 OR BSD-3-Clause)
 
 -- CI script, compatible with all of Travis, Appveyor and Azure.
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -267,10 +308,10 @@ buildDists ghcFlavor resolver = do
         putStrLn $ "# Completed in " ++ showDuration t ++ ": " ++ x ++ "\n"
         hFlush stdout
 
-      mkTarball :: String -> String -> IO ()
```
cocreature: Are you sure that the resolver doesn’t matter here? Seems safer to add it if it doesn’t cause problems.

shayne-fletcher-da: The change here is  that `mkTarball` now calls the function `stack` (rather than `cmd "stack"` directly) so indeed,  the resolver is taken into account.

cocreature: Oh makes sense. Thanks for the explanation!

PR #175, opened 2020-01-17
Add support for ghc-8.8.2

shayne-fletcher:
Add new config for ghc-8.8.2.


shayne-fletcher:
> We already had problems with the number of CI jobs here so I would strongly prefer to just replace `8.8.1` by `8.8.2` instead of testing both.

I didn't know that. Sounds reasonable to me. I'll make it so.

PR #182, opened 2020-02-05
Move GHC.Hs.Dump to ghc-lib-parser

shayne-fletcher:
Allows to use `showDataAst` without requiring ghc-lib.

### DIFF
--- a/examples/strip-locs/src/Main.hs
+++ b/examples/strip-locs/src/Main.hs
@@ -47,9 +47,9 @@ import "ghc-lib-parser" Platform
 #endif
 
 #if defined (GHC_MASTER) || defined (GHC_8101)
-import "ghc-lib" GHC.Hs.Dump
+import "ghc-lib-parser" GHC.Hs.Dump
 #else
-import "ghc-lib" HsDumpAst
+import "ghc-lib-parser" HsDumpAst
 #endif
 
 import Control.Monad
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -295,10 +295,13 @@ calcParserModules ghcFlavor = do
       -- Lastly, manipulate text like 'GHC/Exts/Heap/Constants.hs'
       -- into 'GHC.Exts.Heap.Constants'.
       modules = map (replace "/" "." . dropSuffix ".hs") strippedModulePaths
-  -- We put 'HeaderInfo' here because doing so means clients who
-  -- just want to do parsing don't need 'ghc-lib' (this module
-  -- is needed for parsing in the presence of dynamic pragmas).
-  return $ nubSort (modules ++ ["HeaderInfo"])
+      -- The modules in this list would by default end up in
+      -- ghc-lib. We intervene so that rather, they go into
+      -- ghc-lib-parser.
+      extraModules =
+        ["HeaderInfo"] ++
+        [ if ghcFlavor `elem` [ GhcMaster, Ghc8101 ] then "GHC.Hs.Dump" else "HsDumpAst"]
+  return $ nubSort (modules ++ extraModules)
 
 -- | Selectively disable optimizations in some particular files so as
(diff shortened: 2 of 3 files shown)


examples/strip-locs/src/Main.hs
```diff
@@ -47,9 +47,9 @@ import "ghc-lib-parser" Platform
 #endif
 
 #if defined (GHC_MASTER) || defined (GHC_8101)
-import "ghc-lib" GHC.Hs.Dump
+import "ghc-lib-parser" GHC.Hs.Dump
```
cocreature: Can we change the cabal file of this example to no longer depend on `ghc-lib`?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -295,10 +295,20 @@ calcParserModules ghcFlavor = do
...
       modules = map (replace "/" "." . dropSuffix ".hs") strippedModulePaths
-  -- We put 'HeaderInfo' here because doing so means clients who
-  -- just want to do parsing don't need 'ghc-lib' (this module
-  -- is needed for parsing in the presence of dynamic pragmas).
-  return $ nubSort (modules ++ ["HeaderInfo"])
+      -- The modules in this list would by default end up in
+      -- ghc-lib. We intervene so that rather, they go into
+      -- ghc-lib-parser.
+      extraModules =
+        [ f | True -- Invariant over versions.
```
cocreature:
```suggestion
        [ "HeaderInfo" ]
```
I think trying to making this look like the other cases is just confusing.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -295,10 +295,20 @@ calcParserModules ghcFlavor = do
...
+            , f <- [ "HeaderInfo" ]
+        ] ++
+        [ f | ghcFlavor `elem` [ GhcMaster, Ghc8101 ] -- Modern GHC.
```
cocreature:
```suggestion
        [ "GHC.Hs.Dump" | ghcFlavor `elem` [ GhcMaster, Ghc8101 ] ]
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -295,10 +295,20 @@ calcParserModules ghcFlavor = do
...
+            , f <- [ "GHC.Hs.Dump" ]
+        ] ++
+        [ f | ghcFlavor `notElem` [ GhcMaster, Ghc8101 ] -- Legacy GHC.
```
cocreature:
```suggestion
        [ "HsDumpAst" | ghcFlavor `notElem` [ GhcMaster, Ghc8101 ]]
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -295,10 +295,20 @@ calcParserModules ghcFlavor = do
...
+            , f <- [ "HsDumpAst" ]
```
cocreature: I think rather than `elem` and `notElem` checks you are better off with just an `if` to make it clear that the second case hits when the first doesn’t.

shayne-fletcher: My idea was to build in support for extending those lists but as you think it's overkill, i will simplify

examples/strip-locs/src/Main.hs
shayne-fletcher: Ah yes!

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Yep. Falling back on 
```haskell
      extraModules =
        ["HeaderInfo"] ++
        [ if ghcFlavor `elem` [ GhcMaster, Ghc8101 ] then "GHC.Hs.Dump" else "HsDumpAst"]
```

shayne-fletcher:
> LGTM with a bunch of comments, thanks!

Appreciate it 🤗

PR #188, opened 2020-03-02
Add HeapPrim.cmm to c-sources; remove applyPatchHeapClosures

shayne-fletcher:
It turns out that Cabal will build CMM files if you put them in "c-sources".  In this PR we do this for "libraries/ghc-heap/cbits/HeapPrim.cmm" which enables us to remove `applyPatchHeapClosures` 🎉

Cc @ndmitchell


shayne-fletcher:
The file "libraries/ghc-heap/GHC/Exts/Heap/Closures.hs" requires the definitions provided by "libraries/ghc-heap/cbits/HeapPrim.cmm" in order to compile. We used to satisfy those symbols by providing substituting dummy definitions not because we wanted to but rather because we couldn't work out how to get Cabal to build HeapPrim.cmm. That was what `applyPatchHeapClosures` did. Now we know how to get Cabal to compile CMM files so the motivation is, remove the hack that substitutes for those symbols. 

The two symbols I'm talking about are,
```
#include "Cmm.h"

aToWordzh (P_ clos)
{
    return (clos);
}

reallyUnsafePtrEqualityUpToTag (W_ clos1, W_  clos2)
{
    clos1 = UNTAG(clos1);
    clos2 = UNTAG(clos2);
    return (clos1 == clos2);
}
```

Issue #194, opened 2020-04-12
ghc-lib-8.8.3.* base dependency out-of-bounds error with ghc-8.10 


shayne-fletcher:
See https://github.com/ndmitchell/hlint/issues/951. ghc-8.8.3.20200224 + cabal 3.2.0.0 + ghc-8.10 produces a configure error because the upper-limit of base is 4.14.

shayne-fletcher:
There is also an issue with ghc-prim. See https://github.com/ndmitchell/hlint/issues/951#issuecomment-612638438. That one could be a little fiddlier.

essandess:
Re: https://github.com/ndmitchell/hlint/issues/951#issuecomment-612641551
> @essandess I'm flying blind since I don't have the capacity to build with 8.10 yet (waiting on a stack resolver). If I were to give you a .tar.gz to test, would you be able to do that?

Possibly. I just bootstrapped ghc 8.10.1 from the prebuilt binary, and am using the prebuilt cabal 3.2.0.0 binary. See source sites in:
* https://github.com/macports/macports-ports/blob/master/lang/ghc/Portfile
* https://github.com/macports/macports-ports/blob/master/lang/cabal/Portfile

Would it possible for you to download these binaries, clone hlint, and run a modified version of this build command?```bash
/opt/local/bin/cabal new-configure --prefix=/opt/local --bindir=/opt/local/bin --libdir=/opt/local/lib --libsubdir=hlint --dynlibdir=/opt/local/lib --libexecdir=/opt/local/libexec --libexecsubdir=hlint --datadir=/opt/local/share/hlint --docdir=/opt/local/share/doc/hlint --htmldir=/opt/local/share/doc/hlint --sysconfdir=/opt/local/etc/hlint --enable-documentation --enable-relocatable
```

shayne-fletcher:
Actually I just got a build of ghc-lib-parser with 8.10 using ghc-prim-6.1 so I think we'll be OK. The fiddliness I was worried about doesn't look like it's going to manifest yet (https://gitlab.haskell.org/ghc/ghc/commit/705a16df02411ec2445c9a254396a93cabe559ef). Just running one more very specific test and if that works out, I'll cut a new ghc-lib-parser.

shayne-fletcher:
@essandess ghc-lib-parser-8.8.3.20200412.1 (https://hackage.haskell.org/package/ghc-lib-parser-8.8.3.20200412) just went to hackage now.

PR #196, opened 2020-04-12
Increase upper-bound on ghc-prim

shayne-fletcher:
Closes https://github.com/digital-asset/ghc-lib/issues/194.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -499,7 +499,7 @@ withCommas ms =
 -- | Common build dependencies.
 commonBuildDepends :: [String]
 commonBuildDepends =
-  [ "ghc-prim > 0.2 && < 0.6"
+  [ "ghc-prim > 0.2 && < 0.7"
   , "base >= 4.11 && < 4.15"
   , "containers >= 0.5 && < 0.7"
   , "bytestring >= 0.9 && < 0.11"


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -499,7 +499,7 @@ withCommas ms =
 -- | Common build dependencies.
 commonBuildDepends :: [String]
 commonBuildDepends =
-  [ "ghc-prim > 0.2 && < 0.6"
+  [ "ghc-prim > 0.2 && < 0.6.1"
```
cocreature, line 502: Why `0.6.1` instead of `0.7`? I would expect ghc-prim to uphold pvp.

shayne-fletcher: I'm changing to 0.7 actually as we speak.

shayne-fletcher:
> lgtm but unless there is a reason for choosing 0.6.1, 0.7 seems to make more sense. This also seems like something that we should test in CI instead of simply bumping the bound.

Yes, going with 0.7. Regarding testing, well, this PR tests it in CI. I do agree though that this aspect of the project needs some love.

cocreature:
What I mean is that we should test that we can build ghc-lib against ghc-prim 0.6 in CI. That doesn’t seem to be part of this PR.

shayne-fletcher:
> What I mean is that we should test that we can build ghc-lib against ghc-prim 0.6 in CI. That doesn’t seem to be part of this PR.

Let's come back to it. For now be assured I am checking these facts locally.

shayne-fletcher:
https://github.com/digital-asset/ghc-lib/issues/197

Issue #197, opened 2020-04-12
Construct a CI test with explicit dependencies


shayne-fletcher:
We need a CI test that lists all dependencies to make testing bounds configuration easy.

PR #198, opened 2020-04-13
Add ghc-8.10.1 build support and framework for bounds testing 

shayne-fletcher:
Add support for building/testing with ghc-8.10.1 and a framework for bounds testing. Closes https://github.com/digital-asset/ghc-lib/issues/197.

### DIFF
--- a/stack-exact.yaml
+++ b/stack-exact.yaml
@@ -1,0 +1,64 @@
+# To use this, the command you want is:
+#  stack runhaskell \
+#    --stack-yaml stack-exact.yaml \
+#    --package extra --package optparse-applicative \
+#    CI.hs -- \
+#    --stack-yaml stack-exact.yaml --ghc-flavor xxx
+# You can add --resolver xxx to the above command to have the effet of
+# overriding the choice of compiler if desired.
+
+# Stack 8.10 builds generate a lot of warnings of this kind.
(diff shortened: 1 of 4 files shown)


shayne-fletcher:
Also update to https://gitlab.haskell.org/ghc/ghc.git `e8029816fda7602a8163c4d2703ff02982a3e48c`

stack-exact.yaml
```diff
@@ -0,0 +1,68 @@
...
+- optparse-applicative-0.15.1.0
+- semigroups-0.19.1
+- transformers-compat-0.6.5
+
+# Hack(SF, 2020-04-13) : I don't like this at all but it's the only
+# way I can get a build plan. With this a build plan is constructed
+# that selects transformers-compat-0.6.5 and generates the following
+# warning.
+#   WARNING: Ignoring transformers-compat's bounds on transformers
+#   (>=0.3 && ==0.2.*)
```
cocreature, line 62: I don’t think this is the right solution here. `transformers-compat` is perfectly happy with `transformers-0.5`. The issue here is that it has an automatic cabal flag and Stack does not trigger those. If you enable the `five-three` flag, things should work as expected.

shayne-fletcher: Ah! Of course. Thank-you!

Issue #204, opened 2020-05-02
ghc-lib-parser does not build with GHCJS: 'ghc_boot_platform.h' file not found


neongreen:
```
Preprocessing library for ghc-lib-parser-8.10.1.20200412..
Building library for ghc-lib-parser-8.10.1.20200412..

In file included from compiler/basicTypes/Avail.hs:7:0: error:

/nix/store/23r29nvkgr6xhr9xx3kbxqb313jxmans-ghcjs-8.6.0.1/lib/ghcjs-8.6.0.1/include/HsVersions.h:16:10: error:
     fatal error: 'ghc_boot_platform.h' file not found
   |
16 | #include "ghc_boot_platform.h"
   |          ^
#include "ghc_boot_platform.h"
         ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
`cc' failed in phase `C pre-processor'. (Exit code: 1)
```

neongreen:
I assume this file should be shipped with ghc-lib-parser, similarly to many other `.h` files, but I might be wrong.

shayne-fletcher:
This file ships with `ghc-lib` versions for GHC flavor < 8.10. It is not produced for later versions (in this GHC commit it seems https://gitlab.haskell.org/ghc/ghc/commit/05419e55cab272ed39790695f448b311f22669f7).

shayne-fletcher:
> `ghcjs-8.6.0.1/include/HsVersions.h:16:10: error`

ghc-lib ships with it's own `HsVersions.h`. Why are we pulling in the `ghcjs` one?

shayne-fletcher:
The include dirs from `ghc-lib-parser.cabal` state:
```
    include-dirs:
        includes
        ghc-lib/stage0/lib
        ghc-lib/stage0/compiler/build
        compiler
```
and `compiler/HsVersions.h` is what we should be getting. I have no experience with ghcjs but I get the feeling somehow the include path is getting manipulated in an unexpected way.

neongreen:
I renamed `HsVersions.h` along with references to it, to ensure it doesn't clash with GHC-shipped includes. That was enough for ghc-lib-parser to compile.

Perhaps it should be done for all includes, if there's an easy way to do so in the generator? It might also fix https://github.com/digital-asset/ghc-lib/issues/184:

```
Preprocessing library for ghc-lib-parser-8.8.0.20190424..
In file included from /usr/lib64/ghc-8.6.5/include/Stg.h:319,
```

PR #208, opened 2020-05-18
Add patch HsVersions.h -> GhclibHsVersions.h

shayne-fletcher:
Add a patch step that renames `HsVersions.h` to `GhclibHsVersions.h`. This enables ghc-lib-parser to be used in conjunction with ghcjs. Fixes https://github.com/digital-asset/ghc-lib/issues/204.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -3,7 +3,8 @@
 -- (Apache-2.0 OR BSD-3-Clause)
 
 module Ghclibgen (
-    applyPatchGhcPrim
+    applyPatchHsVersions
+  , applyPatchGhcPrim
   , applyPatchDisableCompileTimeOptimizations
   , applyPatchRtsIncludePaths
   , applyPatchStage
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -326,7 +328,23 @@ calcParserModules ghcFlavor = do
...
 
--- | Selectively disable optimizations in some particular files so as
+-- Rename 'HsVersions.h' to 'GhclibHsVersions.h' then replace
+-- occurences of that string in all .hs and .y files reachable from
+-- the 'compiler' directory. This enables use of ghc-lib-parser with
+-- ghcjs (issue https://github.com/digital-asset/ghc-lib/issues/204).
+applyPatchHsVersions :: GhcFlavor -> IO ()
+applyPatchHsVersions _ = do
+  renameFile "compiler/HsVersions.h" "compiler/GhclibHsVersions.h"
+  files <- find always (extension ==? ".hs" ||? extension ==? ".y") "compiler"
```
cocreature, line 338: `listFilesRecursive` + `filter` seems simpler and avoids the extra dependency.

shayne-fletcher: Ah yes. Good catch.

PR #209, opened 2020-05-20
store da-ghc-lib on a GCS bucket

garyverhaegen-da:
At the moment, publishing a new version of da-ghc-lib (the one DAML depends on) is a manual process aimed at Bintray. As we have just received a note from JFrog that Bintray is sunsetting, this is a good opportunity to automate the deployment.

This PR aims at creating a GCS bucket (Terraform files to be applied manually) and change the CI process to push every master commit to said GCS bucket.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,143 +16,173 @@ pr:
     include:
     - master
 
-strategy:
-  # Limit number of executors used so other pipelines can run too
-  maxParallel: 10
-  matrix:
-    # Notes:
-    #  Tags are encoded in the following way:
-    #        <os> '-' <ghc-lib> '-' <compiler>
-    #  Not every combination is tested
-    #  - We do sampling to keep the number of builds reasonable;
-    #  - No ghc-8.4.4 builds are tested;
-    #  - No Windows 8.8.3 builds are tested (ghc-8.8.3 on Windows
-    #    has issues);
-    #  - Minimum GHC needed to bootstrap:
-    #      +--------------+------------+
-    #      | GHC flavor   |  version   |
-    #      +==============+============+
-    #      | ghc-8.8.*    |  >= 8.4.4  |
-    #      | ghc-8.10.*   |  >= 8.6.5  | (since 09/29/2019, https://gitlab.haskell.org/ghc/ghc/commit/24620182abdfcc65a0cfc0e2f3bc391d464d0ad5)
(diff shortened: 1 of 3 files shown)


garyverhaegen-da:
Note: as of opening this PR, there are a number of missing pieces:

- I will need some help to complete the build/upload process. I have left `???` markers where I think I need help, but feel free to correct anything else too.
- I have not yet applied the Terraform files (waiting for @hurryabit's approval for that), so the bucket doesn't exist yet. This in turn means the CI writer credentials don't exist yet either and are therefore not yet added to Azure.

Additional notes:

- The `azure-pipelines.yml` file was written from the perspective of a single job; I had to add a separate job, which means indenting all of the existing content. The diff on this PR is therefore better viewed [without whitespace diffs](https://github.com/digital-asset/ghc-lib/pull/209/files?w=1).
- We do not have a process for tracking infra changes from this repo for future audits. I expect this to be a one-off, so my plan for now is to create a corresponding Standard-Change PR on the daml repo (with an empty commit and a link to this PR) so it will appear in our monthly report. Please advise if this doesn't seem right.

garyverhaegen-da:
Oh and also as usual once this is all done and applied and merged we'll still need Ed to create the DNS entry before the bucket is actually useable.

shayne-fletcher:
> * Not sure about the calculation of `VERSION_TAG` - will it match what `CI.hs` generates for the tarball name(s)? What happens if you find yourself needing to do a second one on the same day? e.g. ghc-lib-8.10.1.20200518.1.tar.gz

Could be a small enhancement to CI.hs might help here (maybe, maybe not). I've implemented a `--version-tag` option in ghc-lib-parser-ex that looks like it might be of use here in that we could use that feature to "inject" the desired version name into the build process from the script exogenously.

garyverhaegen-da:
This is the wrong repo for this, and probably the wrong approach. Thanks @shayne-fletcher for all the additional context; closing this for now.

cocreature:
I’m not sure this is necessarily the wrong repo. It depends on both. Right now, there is no CI on the ghc repo at all so adding it here seems like a reasonable option. We could just have a single file that locks down the revision and if that changes we publish.

shayne-fletcher:
> I’m not sure this is necessarily the wrong repo. It depends on both. Right now, there is no CI on the ghc repo at all so adding it here seems like a reasonable option. We could just have a single file that locks down the revision and if that changes we publish.

Indeed. Gary and I ended up discussing that scheme and it may well be what gets done. He's mulling it over.

garyverhaegen-da:
For anyone stumbling on this in the future and finding the suspense unbearable, this has been implemented across both this and the daml repo as https://github.com/digital-asset/daml/pull/6188 and https://github.com/digital-asset/ghc-lib/pull/215.

Issue #210, opened 2020-05-23
Fail to build with stack + ghc-8.6.5 + windows


jneira:
Hi, first of all congrats for your work with this library.
We are getting an error for the mentioned configuration (and only for that one, cabal builds it succesfully):

```
haskell-language-server> GHC runtime linker: fatal error: I found a duplicate definition for symbol
haskell-language-server>    aToWordzh
haskell-language-server> whilst processing object file
haskell-language-server>    D:\bin\stack\x86_64-windows\ghc-8.6.5\lib\ghc-heap-8.6.5\HSghc-heap-8.6.5.o
haskell-language-server> The symbol was previously defined in
haskell-language-server>    C:\sr\snapshots\9070fc3e\lib\x86_64-windows-ghc-8.6.5\ghc-lib-parser-8.10.1.20200412-2BSV0WaFUv7bn4MuxkBnD\HSghc-lib-parser-8.10.1.20200412-2BSV0WaFUv7bn4MuxkBnD.o
haskell-language-server> This could be caused by:
haskell-language-server>    * Loading two different object files which export the same symbol
haskell-language-server>    * Specifying the same object file twice on the GHCi command line
haskell-language-server>    * An incorrect `package.conf' entry, causing some object to be
haskell-language-server>      loaded twice.
haskell-language-server> ghc.exe: panic! (the 'impossible' happened)
haskell-language-server>   (GHC version 8.6.5 for x86_64-unknown-mingw32):
haskell-language-server>        loadObj "D:\\bin\\stack\\x86_64-windows\\ghc-8.6.5\\lib\\ghc-heap-8.6.5\\HSghc-heap-8.6.5.o": failed
haskell-language-server>
haskell-language-server> Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
haskell-language-server>

--  While building package haskell-language-server-0.1.0.0 using:
      C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_Z6RU0evB_2.4.0.1_ghc-8.6.5.exe --builddir=.stack-work\dist\e626a42b build 
lib:haskell-language-server lib:hls-test-utils exe:haskell-language-server exe:haskell-language-server-wrapper --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
)
```

The original issue in haskell-language-server is: https://github.com/haskell/haskell-language-server/issues/90
Same for version `ghc-lib-parser-8.10.1.20200518`

[...]

PR #212, opened 2020-05-23
Obfuscate ghc heap symbols

shayne-fletcher:
Duplicate symbol errors arise on Windows when linking with ghc-lib-parser with ghc-heap. Syntactically obfuscate ghc-lib's `aToWordzh` and `reallyUnsafePtrEqualityUpToTag` to avoid this.  Fixes https://github.com/digital-asset/ghc-lib/issues/210.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -3,7 +3,8 @@
 -- (Apache-2.0 OR BSD-3-Clause)
 
 module Ghclibgen (
-    applyPatchHsVersions
+    applyPatchHeapClosures
+  , applyPatchHsVersions
   , applyPatchGhcPrim
   , applyPatchDisableCompileTimeOptimizations
   , applyPatchRtsIncludePaths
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
This observation is likely a knock-on effect of https://github.com/digital-asset/ghc-lib/pull/188.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -327,6 +328,27 @@ calcParserModules ghcFlavor = do
...
+    replace
+      "reallyUnsafePtrEqualityUpToTag"
+      "Ghclib_reallyUnsafePtrEqualityUpToTag"
+    =<< readFile' "libraries/ghc-heap/cbits/HeapPrim.cmm"
+  writeFile "libraries/ghc-heap/GHC/Exts/Heap/Closures.hs" .
+    replace
+      "\"aToWordzh\""
+      "\"Ghclib_aToWordzh\"" .
+    replace
+      "\"reallyUnsafePtrEqualityUpToTag\""
```
cocreature, line 348: Do we need the quotes here? If it really was ambiguous without the quotes the other replacement is also dangerous so I’m not sure it helps much.

shayne-fletcher: The phrasing makes sure `reallyUnsafePtrEqualityUpToTag` in Closures.hs is replaced but not `reallyUnsafePtrEqualityUpToTag#`. I didn't test not doing it that way (expecting it to fail) and put my focus instead on carefully checking what had to happen and verifying the result. In HeapPrim.cmm there was no chance of the substitutions going wrong in that fashion and thus would be correct applied unconditionally.

PR #215, opened 2020-06-02
more flexible da flavour

garyverhaegen-da:
This PR updates the `CI.hs` script to make the DAML-specific build more flexible. The motivation for this change is to set up a fully auditable CI process for the DAML repo, i.e. one in which we can specify (and track) exactly which commits were involved in creating the version of ghc-lib used by the DAML compiler.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -28,7 +28,7 @@ main = do
     let opts =
           Opts.info (parseOptions Opts.<**> Opts.helper)
           ( Opts.fullDesc
-            <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
+            <> Opts.progDesc "Build ghc-lib and ghc-lib-parser tarballs."
             <> Opts.header "CI - CI script for ghc-lib"
           )
     Options { ghcFlavor, stackOptions } <- Opts.execParser opts
@@ -126,20 +126,20 @@ parseOptions = Options
    parseDaOptions =
        Opts.flag' Da ( Opts.long "da" <> Opts.help "Enables DA custom build." )
        <*> Opts.strOption
-           ( Opts.long "da-merge-base-sha"
+           ( Opts.long "merge-base-sha"
           <> Opts.help "DA flavour only. Base commit to use from the GHC repo."
           <> Opts.showDefault
           <> Opts.value "ghc-8.8.1-release"
            )
        <*> (Opts.some
              (Opts.strOption
-              ( Opts.long "da-patch"
-             <> Opts.help "DA flavour only. Commits to merge in from the DA GHC fork, referenced as 'upstream'. Can be specified multiple times. If no patch is specified, default will be equivalent to `--da-patch upstream/da-master-8.8.1 --da-patch upstream/da-unit-ids-8.8.1`. Specifying any patch will overwrite the default (i.e. replace, not add)."
+              ( Opts.long "patch"
+             <> Opts.help "DA flavour only. Commits to merge in from the DA GHC fork, referenced as 'upstream'. Can be specified multiple times. If no patch is specified, default will be equivalent to `--patch upstream/da-master-8.8.1 --patch upstream/da-unit-ids-8.8.1`. Specifying any patch will overwrite the default (i.e. replace, not add)."
               ))
             Opts.<|>
             pure ["upstream/da-master-8.8.1", "upstream/da-unit-ids-8.8.1"])
        <*> Opts.strOption
-           ( Opts.long "da-gen-flavor"
+           ( Opts.long "gen-flavor"
           <> Opts.help "DA flavor only. Flavor to pass on to ghc-lib-gen."
           <> Opts.showDefault
           <> Opts.value "da-ghc-8.8.1")
--- a/CI.hs
+++ b/CI.hs
@@ -31,15 +31,12 @@ main = do
             <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
             <> Opts.header "CI - CI script for ghc-lib"
           )
-    Options { ghcFlavor, daMergeBaseSha, daPatches, daFlavor, stackOptions } <- Opts.execParser opts
-    version <- buildDists ghcFlavor stackOptions daMergeBaseSha daPatches daFlavor
+    Options { ghcFlavor, stackOptions } <- Opts.execParser opts
+    version <- buildDists ghcFlavor stackOptions
     putStrLn version
 
 data Options = Options
     { ghcFlavor :: GhcFlavor
-    , daMergeBaseSha :: String
-    , daPatches :: [String]
-    , daFlavor :: String
     , stackOptions :: StackOptions
     } deriving (Show)
 
@@ -51,7 +48,15 @@ data StackOptions = StackOptions
     , ghcOptions :: Maybe String -- If 'Just _', pass '--ghc-options="xxx"' to 'stack build' (for ghc verbose, try 'v3').
     } deriving (Show)
 
-data GhcFlavor = Ghc8101 | Ghc881 | Ghc882 | Ghc883 | Da | GhcMaster String
+data GhcFlavor = Ghc8101
+               | Ghc881
+               | Ghc882
+               | Ghc883
+               | Da { mergeBaseSha :: String
+                    , patches :: [String]
+                    , flavor :: String
+                    }
+               | GhcMaster String
   deriving (Eq, Show)
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
@@ -70,13 +75,13 @@ stackResolverOpt = \case
   Just resolver -> "--resolver " ++ resolver
   Nothing -> ""
 
-ghcFlavorOpt :: String -> GhcFlavor -> String
-ghcFlavorOpt daFlavor = \case
+ghcFlavorOpt :: GhcFlavor -> String
+ghcFlavorOpt = \case
     Ghc8101 -> "--ghc-flavor ghc-8.10.1"
     Ghc881 -> "--ghc-flavor ghc-8.8.1"
     Ghc882 -> "--ghc-flavor ghc-8.8.2"
     Ghc883 -> "--ghc-flavor ghc-8.8.3"
-    Da -> "--ghc-flavor " <> daFlavor
+    Da { flavor } -> "--ghc-flavor " <> flavor
     GhcMaster _hash -> "--ghc-flavor ghc-master"
       -- The git SHA1 hash is not passed to ghc-lib-gen at this time.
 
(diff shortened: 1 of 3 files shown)


garyverhaegen-da:
This should be reviewed in the context of https://github.com/digital-asset/daml/pull/6188.

CI.hs
```diff
@@ -101,13 +103,28 @@ genVersionStr = \case
...
-    <$> Opts.switch
-        ( Opts.long "upload-to-bintray"
-       <> Opts.help "If specified, will try uploading the sdists to Bintray, using credentials in BINTRAY_BASIC_AUTH env var.")
-    <*> Opts.option readFlavor
+    <$> Opts.option readFlavor
         ( Opts.long "ghc-flavor"
        <> Opts.help "The ghc-flavor to test against"
         )
+    <*> Opts.strOption
+        ( Opts.long "da-base-sha"
```
cocreature, line 111: Maybe add a `ghc` in there and potentially `upstream` might be clearer than `base`.

CI.hs
```diff
@@ -101,13 +103,28 @@ genVersionStr = \case
...
+       <> Opts.help "DA flavour only. Base commit to use from the GHC repo."
+       <> Opts.showDefault
+       <> Opts.value "ghc-8.8.1-release"
```
cocreature, line 114: Did you just create this tag? If not, I don’t understand what this is pointing to.

CI.hs
```diff
@@ -101,13 +103,28 @@ genVersionStr = \case
...
+        )
+    <*> Opts.strOption
```
cocreature, line 116: Is there a reason why you don’t allow passing this multiple times one per patch instead of accepting a space-separated option? The latter seems a bit weird but I don’t have strong feelings here so happy to go with your preference.

CI.hs
garyverhaegen-da: Not my addition, we were already checking out that tag. As far as I can tell it's [the official release tag](https://gitlab.haskell.org/ghc/ghc/-/tree/ghc-8.8.1-release) on the GHC repo.

CI.hs
garyverhaegen-da: I had not thought of that. I'll have a look at how to do that with optparse.

CI.hs
garyverhaegen-da: It's the base sha in the sense that it's the one we merge on to. Calling it upstream here would probably be confusing as we currently add our own fork as `upstream`, and this tag comes from `origin`.

CI.hs
cocreature: Makes sense, I got a bit confused and thought this was something we created.

CI.hs
cocreature: :+1:

CI.hs
cocreature: `many` is the function you need for this.

garyverhaegen-da: Switched to `some`, because I wanted a default value and `many` loops if there is one.

CI.hs
garyverhaegen-da: Renamed to `da-merge-base-sha`, hopefully that will be clearer.

CI.hs
```diff
@@ -30,13 +31,15 @@ main = do
             <> Opts.progDesc "Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs."
             <> Opts.header "CI - CI script for ghc-lib"
           )
-    Options { upload, ghcFlavor, stackOptions } <- Opts.execParser opts
-    version <- buildDists ghcFlavor stackOptions
-    when upload $ bintrayUpload version
+    Options { ghcFlavor, daMergeBaseSha, daPatches, daFlavor, stackOptions } <- Opts.execParser opts
```
shayne-fletcher, line 34: At the least I'd like to see the DA options bundled into a structure ala `stackOptions`. I think that structure should be an optional only defined in the presence of a DA build rather than always defined with defaults.

CI.hs
```diff
@@ -48,7 +51,7 @@ data StackOptions = StackOptions
     , ghcOptions :: Maybe String -- If 'Just _', pass '--ghc-options="xxx"' to 'stack build' (for ghc verbose, try 'v3').
     } deriving (Show)
 
-data GhcFlavor = Ghc8101 | Ghc881 | Ghc882 | Ghc883 | DaGhc881 | GhcMaster String
+data GhcFlavor = Ghc8101 | Ghc881 | Ghc882 | Ghc883 | Da | GhcMaster String
```
shayne-fletcher, line 54: Ideally I'd like to see the DA specific parameters on the `Da` ctor as we do for the `GhcMaster` ctor. That `String` you see there is the commit SHA for GHC in the case of flavor `GhcMaster XXX`. Better we have one way of doing things so I'd rather see this mechanism extended if possible.

CI.hs
cocreature: I agree but let’s please do that separately.

CI.hs
```diff
@@ -51,7 +48,15 @@ data StackOptions = StackOptions
     , ghcOptions :: Maybe String -- If 'Just _', pass '--ghc-options="xxx"' to 'stack build' (for ghc verbose, try 'v3').
     } deriving (Show)
 
-data GhcFlavor = Ghc8101 | Ghc881 | Ghc882 | Ghc883 | Da | GhcMaster String
+data GhcFlavor = Ghc8101
+               | Ghc881
+               | Ghc882
+               | Ghc883
+               | Da { mergeBaseSha :: String
```
shayne-fletcher, line 55: Shouldn't comment on style matters but I think this will fit on a single line eh?

CI.hs
```diff
@@ -70,13 +75,13 @@ stackResolverOpt = \case
...
-ghcFlavorOpt :: String -> GhcFlavor -> String
-ghcFlavorOpt daFlavor = \case
+ghcFlavorOpt :: GhcFlavor -> String
+ghcFlavorOpt = \case
     Ghc8101 -> "--ghc-flavor ghc-8.10.1"
     Ghc881 -> "--ghc-flavor ghc-8.8.1"
     Ghc882 -> "--ghc-flavor ghc-8.8.2"
     Ghc883 -> "--ghc-flavor ghc-8.8.3"
-    Da -> "--ghc-flavor " <> daFlavor
+    Da { flavor } -> "--ghc-flavor " <> flavor
```
shayne-fletcher, line 84: This is good. That is the flag that gets pushed "downstream" into `Ghclibgen`. I like it.

CI.hs
```diff
@@ -135,9 +123,29 @@ parseOptions = Options
        "ghc-8.8.1" -> Right Ghc881
        "ghc-8.8.2" -> Right Ghc882
        "ghc-8.8.3" -> Right Ghc883
-       "da" -> Right Da
        "ghc-master" -> Right (GhcMaster current)
        hash -> Right (GhcMaster hash)
+   parseDaOptions :: Opts.Parser GhcFlavor
+   parseDaOptions =
+       Opts.flag' Da ( Opts.long "da" <> Opts.help "Enables DA custom build." )
```
shayne-fletcher, line 130: Maybe `da-ghc-flavor`? Something to think about. I guess this has to be exclusive with respect to `ghc-flavor`. Does the parser enforce that?

CI.hs
```diff
@@ -197,12 +202,12 @@ buildDists
         Ghc881 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.1-release"
         Ghc882 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.2-release"
         Ghc883 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.3-release"
-        Da -> do
-            cmd $ "cd ghc && git fetch --tags && git checkout " <> daMergeBaseSha
+        Da { mergeBaseSha, patches } -> do
```
shayne-fletcher, line 205: Exactly what I had in mind! 🎉

CI.hs
garyverhaegen-da: That is my understanding of the [documentation](https://github.com/pcapriotti/optparse-applicative#alternative), and seems to be confirmed in usage:
```
$ stack runhaskell --package extra --package optparse-applicative CI.hs -- --da --ghc-flavor "ghc-8.8.1"
Invalid option `--ghc-flavor'

Usage: CI.hs (--da [--da-merge-base-sha ARG] [--da-patch ARG] 
               [--da-gen-flavor ARG] |
               --ghc-flavor ARG) [--stack-yaml ARG] [--resolver ARG] 
             [--verbosity ARG] [--cabal-verbose] [--ghc-options ARG]
  Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs.
$ stack runhaskell --package extra --package optparse-applicative CI.hs -- --ghc-flavor "ghc-8.8.1" --da
Invalid option `--da'

Usage: CI.hs (--da [--da-merge-base-sha ARG] [--da-patch ARG] 
               [--da-gen-flavor ARG] |
               --ghc-flavor ARG) [--stack-yaml ARG] [--resolver ARG] 
             [--verbosity ARG] [--cabal-verbose] [--ghc-options ARG]
  Build (and possibly upload) ghc-lib and ghc-lib-parser tarballs.
$
```

CI.hs
garyverhaegen-da: I'm fine with whatever style you prefer on this project. I've updated this one (though you need longer-than-80-chars lines for it to fit ;-) ); don't hesitate to point out others. In particular, I have no idea how the `parseOptions` function "should" be indented.

CI.hs
garyverhaegen-da: Which makes me think maybe the other flags don't need the `da` prefix? 🤔  This may be veering close to bikeshedding territory though. Happy to go with whatever you prefer for all flag names.

shayne-fletcher: If they don't need the `da` prefix, let's loose them as an elegance.

CI.hs
shayne-fletcher: "Consistency is the hobgoblin of little minds" 🤣

CI.hs
shayne-fletcher: There's no "upload" facilities any more right? Can the help comment be updated accordingly please?

garyverhaegen-da: Oh right.

README.md
```diff
@@ -49,3 +49,38 @@ stack runhaskell --package extra --package optparse-applicative CI.hs -- --ghc-f
...
+- Add the [DA fork](https://github.com/digital-asset/ghc/) of GHC as a remote
+  named `upstream`.
+- Checkout the commit provided as `merge-base-sha`.
+- Create a new commit by merging in all of the commits specified through the
+  `--patch` flags.
+- Proceed as normal for the rest of the workflow.
+
+At some later stage, the workflow calls out to the `ghc-lib-gen` program, and
+at that point it needs to pass in a "flavor" argument; it will use the value of
+the `--gen-flavor` option for that.
```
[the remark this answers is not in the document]
garyverhaegen-da: In the `--da` case the value passed in is the `--gen-flavor` flag, as there is no `--ghc-flavor`.

PR #218, opened 2020-06-19
Delete ChangeLog.md

maralorn:
It‘s obviously not used.


cocreature:
It looks like this file is referenced from `ghc-lib-gen.cabal`. Can you please also remove it there?

PR #221, opened 2020-06-22
ghc-lib-gen, pure refactorings

shayne-fletcher:
Some refactoring in ghc-lib-gen aimed at cutting down on repetition and syntactic verbosity:
  - Named constants for frequently used paths;
  - Consistent list syntax convention;
  - Use predicates for `ghcFlavor` checks.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -42,24 +42,17 @@ cabalFileLibraries =
     , "compiler/ghc.cabal"
     ]
 
--- These predicates are used so ubiquitously it makes sense to
--- introduce some sugar.
-isGhcMaster, isGhc808, isGhcGreaterEqual810 :: GhcFlavor -> Bool
-isGhcMaster ghcFlavor = ghcFlavor == GhcMaster
-isGhc808 ghcFlavor = ghcFlavor `notElem` [ GhcMaster, Ghc8101]
-isGhcGreaterEqual810 ghcFlavor = ghcFlavor `elem` [ GhcMaster, Ghc8101 ]
-
 -- | C-preprocessor "include dirs" for 'ghc-lib-parser'.
 ghcLibParserIncludeDirs :: GhcFlavor -> [FilePath]
 ghcLibParserIncludeDirs ghcFlavor =
   [ "includes" ] ++ -- ghcconfig.h, MachDeps.h, MachRegs.h, CodeGen.Platform.hs
-  (case ghcFlavor of
-    GhcMaster -> [stage0Lib]
-    Ghc8101   -> [stage0Lib]
-    _         ->  ["ghc-lib/generated"]
+  ( case ghcFlavor of
+      GhcMaster -> [ stage0Lib ]
+      Ghc8101 -> [ stage0Lib ]
+      _ -> [ "ghc-lib/generated" ]
   ) ++
   [ stage0Compiler, "compiler"] ++
-  ["compiler/utils" | isGhc808 ghcFlavor]
+  [ "compiler/utils" | ghcFlavor `elem` [ Ghc881, Ghc882, Ghc883, DaGhc881 ] ]
 
 -- Sort by length so the longest paths are at the front. We do this
 -- so that in 'calcParserModules', longer substituions are performed
@@ -92,12 +85,12 @@ ghcLibParserHsSrcDirs forParserDepends ghcFlavor lib =
         -- This next conditional just smooths over a 'master'
         -- vs. '8.8.1' branch difference (relating to a file
         -- 'GhcVersion.hs' IIRC).
-        [ stage0GhcBoot | isGhcGreaterEqual810 ghcFlavor ] ++
+        [ stage0GhcBoot | ghcFlavor `elem` [ GhcMaster, Ghc8101 ] ] ++
         map takeDirectory cabalFileLibraries ++
         askFiles lib "hs-source-dirs:"
 
       excludes = Set.fromList $
-        map ("compiler/" ++) (
+        map ("compiler" </>) (
           [ "codeGen"
           , "hieFile"
           , "llvmGen"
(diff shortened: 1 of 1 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -42,19 +42,24 @@ cabalFileLibraries =
     , "compiler/ghc.cabal"
     ]
 
+-- These predicates are used so ubiquitously it makes sense to
+-- introduce some sugar.
+isGhcMaster, isGhc808, isGhcGreaterEqual810 :: GhcFlavor -> Bool
+isGhcMaster ghcFlavor = ghcFlavor == GhcMaster
+isGhc808 ghcFlavor = ghcFlavor `notElem` [ GhcMaster, Ghc8101]
```
cocreature, line 49: That’s a weird way to phrase this and will break as soon as we add a new version. Why not check for the versions you want instead of excluding the ones you don’t want?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -101,14 +103,11 @@ ghcLibParserHsSrcDirs forParserDepends ghcFlavor lib =
...
-          [ d | ghcFlavor == GhcMaster
-              , d <- [ "typecheck"
-                     , "specialise"
-                     , "cmm" ]
+          , "stranal"
           ] ++
-          [ "nativeGen" | ghcFlavor /= GhcMaster] ++ -- Since 2020-01-04. See https://gitlab.haskell.org/ghc/ghc/commit/d561c8f6244f8280a2483e8753c38e39d34c1f01.
-          [ "deSugar"   | ghcFlavor `elem` [GhcMaster, Ghc8101] && not forParserDepends]
+          [ d | isGhcMaster ghcFlavor, d <- [ "typecheck", "specialise", "cmm" ] ] ++
+          [ "nativeGen" | (not . isGhcMaster) $ ghcFlavor ] ++ -- Since 2020-01-04. See https://gitlab.haskell.org/ghc/ghc/commit/d561c8f6244f8280a2483e8753c38e39d34c1f01.
```
cocreature, line 109: That seems harder to read and is longer than the previous code.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -42,19 +42,24 @@ cabalFileLibraries =
     , "compiler/ghc.cabal"
     ]
 
+-- These predicates are used so ubiquitously it makes sense to
+-- introduce some sugar.
+isGhcMaster, isGhc808, isGhcGreaterEqual810 :: GhcFlavor -> Bool
+isGhcMaster ghcFlavor = ghcFlavor == GhcMaster
+isGhc808 ghcFlavor = ghcFlavor `notElem` [ GhcMaster, Ghc8101]
+isGhcGreaterEqual810 ghcFlavor = ghcFlavor `elem` [ GhcMaster, Ghc8101 ]
```
cocreature, line 50: Would it make sense to define an `Ord` instance?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -162,33 +171,32 @@ dataFiles =
...
 
 derivedConstantsDependencies :: GhcFlavor -> [FilePath]
 derivedConstantsDependencies ghcFlavor =
-   map (root ++)
-   [ "DerivedConstants.h"
-   , "GHCConstantsHaskellExports.hs"
-   , "GHCConstantsHaskellType.hs"
-   , "GHCConstantsHaskellWrappers.hs"
-   ]
+   map (root </>)
```
cocreature, line 184: Do you really want `(</>)` here?  This ends up in the cabal file so your cabal file is now platform dependent.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -162,33 +171,32 @@ dataFiles =
...
+      root =
+        case ghcFlavor of
+          GhcMaster -> stage0Lib
+          Ghc8101 -> stage0Lib
+          _ -> ghcLibGeneratedPath
 
 compilerDependencies :: GhcFlavor -> [FilePath]
 compilerDependencies ghcFlavor =
-  map ("ghc-lib/stage0/compiler/build/" ++) $
+  map (stage0Compiler </>) $
```
cocreature, line 199: Same here.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -205,19 +213,20 @@ compilerDependencies ghcFlavor =
     , "primop-vector-tys.hs-incl"
     , "primop-vector-uniques.hs-incl"
     ] ++
-    [ "primop-docs.hs-incl" | ghcFlavor == GhcMaster ]
+    [ "primop-docs.hs-incl" | isGhcMaster ghcFlavor ]
```
cocreature, line 216: While this is not longer, I still don’t find it any easier to read.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: That's a great idea.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: `import System.FilePath.Posix((</>)) -- Make sure we generate / on all platforms.` probably means it's safe but I take your point. i'll fix that.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: yep

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: ok. let's get the ord thing going and not mess with the `==` cases.

ghc-lib-gen/src/Ghclibgen.hs
cocreature: ah didn’t realize that. Then it’s fine :+1:

ghc-lib-gen/src/Ghclibgen.hs
cocreature: Note that `Ord` is a bit tricky here. You’ll also want to override `Eq` and then you end up with `Da` and `Ghc881` being identical which could cause problems if you want to differentiate them via `Eq` (not sure we do). If you want to avoid that go for a custom comparison function.

shayne-fletcher: As we concluded off ticket, we'll hold off on this for now and remove the sugar.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: We agree to not change this and leave it as `==`.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Yes. Improved occurrences to be explicit in this fashion.

shayne-fletcher:
I think with what we do accept in this PR it's a teensy bit better. Hope it goes towards the boy scout's motto : "always leave it better than you found it".

shayne-fletcher:
> Nice, thank you!
> Sorry for the back and forth.

Not at all! This is what it's all about. I ❤️  the back and forth. At the end of the day we get a better program!

PR #229, opened 2020-07-24
Sync head : hscTarget, derived constants

shayne-fletcher:
Sync to https://gitlab.haskell.org/ghc/ghc.git `dff1cb3d9c111808fec60190747272b973547c52`.
- Update `hscTarget` to account for https://gitlab.haskell.org/ghc/ghc/-/commit/f7cc431341e5b5b31758eecc8504cae8b2390c10.

### DIFF
--- a/examples/mini-compile/src/Main.hs
+++ b/examples/mini-compile/src/Main.hs
@@ -82,7 +82,11 @@ mkDynFlags filename s = do
   let baseFlags =
         (defaultDynFlags fakeSettings fakeLlvmConfig) {
           ghcLink = NoLink
+#if defined(GHC_MASTER)
+        , backend = NoBackend
+#else
         , hscTarget = HscNothing
+#endif
 #if defined(GHC_MASTER)
         , unitDatabases = Just []
 #else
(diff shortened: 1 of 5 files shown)


shayne-fletcher:
Sync to https://gitlab.haskell.org/ghc/ghc.git `fc0f6fbcd95f2dc69a8efabbee2d8a485c34cc47`
- Change derived constants and compiler dependencies files to adapt to MR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3666

examples/mini-compile/src/Main.hs
```diff
@@ -82,7 +82,11 @@ mkDynFlags filename s = do
   let baseFlags =
         (defaultDynFlags fakeSettings fakeLlvmConfig) {
           ghcLink = NoLink
+#if defined(GHC_MASTER)
```
cocreature, line 85: You need more CPP here judging from CI. `platormMini_os` is no longer a thing.

shayne-fletcher: No it's not. That's a surprise. How are my tests passing locally? I'll look into it.

shayne-fletcher: I have the fix to hand. I just need to see if fail here first 😠

shayne-fletcher:
One more push just now (fixing formatting). Sorry about that. Should be done now.

shayne-fletcher:
> /azp run

🙇

PR #233, opened 2020-08-08
Support for ghc-8.10.2

shayne-fletcher:
Add support for ghc-8.10.2


shayne-fletcher:
- Looks like we will need to wait on a stack resolver `ghc-8.10.2` for this to pass CI.

PR #236, opened 2020-08-14
Sync head; fix default extensions; rename homeUnitId

shayne-fletcher:
- Sync https://gitlab.haskell.org/ghc/ghc.git `8a51b2ab7433c06bddca9699b0dfd8ab1d11879b`;
- Stop assuming the `default-extensions` field is static and populate it properly;
- `homeUnitId` -> `homeUnitId_`.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -657,7 +657,6 @@ generateGhcLibCabal ghcFlavor = do
         , ""
         , "library"
         , "    default-language:   Haskell2010"
-        , "    default-extensions: NoImplicitPrelude"
         , "    exposed: False"
         , "    include-dirs:"
         ] ++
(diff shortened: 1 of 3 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -744,9 +741,8 @@ generateGhcLibParserCabal ghcFlavor = do
         , "    build-depends:"
         ] ++ indent2 (withCommas (commonBuildDepends ghcFlavor)) ++
         [ "    build-tools: alex >= 3.1, happy >= 1.19.4"
-        , "    other-extensions:"
-        ] ++
-        indent2 (askField lib "other-extensions:") ++
+        , "    other-extensions:" ] ++ indent2 (askField lib "other-extensions:") ++
+        [ "    default-extensions:" ] ++ indent2 (askField lib "default-extensions:") ++
```
hurryabit, line 745: This looks like we have the same code twice. Is there a specific reason for this?

shayne-fletcher: Two functions for two cabal files : `ghc-lib-parser.cabal` and `ghc-lib.cabal`. No deeper reason than that.

shayne-fletcher: Unless I misunderstand your question and you are confusing lines 744 and 745 as being the same? (They're not they differ in field-name).

hurryabit: Your first answer answers my question. Thanks a lot.

PR #238, opened 2020-09-01
Sync head

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git `a8a2568b7b64e5b9fca5b12df7da759de4db39ae`;
- Remove some redundant imports from `CI.hs`.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -17,11 +17,8 @@ import Data.Maybe
 import Data.List.Extra
 import Data.Time.Clock
 import Data.Time.Calendar
-import Data.Semigroup ((<>))
 import qualified Data.List
 import qualified Options.Applicative as Opts
-import qualified System.Environment as Env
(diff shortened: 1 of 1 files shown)


CI.hs
```diff
@@ -17,11 +17,8 @@ import Data.Maybe
 import Data.List.Extra
 import Data.Time.Clock
 import Data.Time.Calendar
-import Data.Semigroup ((<>))
```
cocreature, line 20: Maybe time to test this on CI with `-Werror` but let’s leave that for another PR.

cocreature (approved):
Thanks!

CI.hs
shayne-fletcher: PR https://github.com/digital-asset/ghc-lib/pull/240 should deal with it (requires rebasing on this).

Issue #243, opened 2020-09-16
 ghc-lib-8.10.* fails to build with "Not in scope: ‘Prelude.*’"


jneira:
* os: windows
* ghc-8.8.4
* cabal-3.2.0.0
* log of ` cabal get ghc-lib-8.10.2.20200808 && cd ghc-lib-8.10.2.20200808 && cabal build`:

```
Resolving dependencies...
Build profile: -w ghc-8.8.4 -O1
In order, the following will be built (use -v for more details):
 - happy-1.20.0 (exe:happy) (requires build)
 - ghc-lib-parser-8.10.2.20200808 (lib) (requires build)
 - ghc-lib-8.10.2.20200808 (lib) (first run)
Configuring executable 'happy' for happy-1.20.0..
Preprocessing executable 'happy' for happy-1.20.0..
....
Configuring library for ghc-lib-parser-8.10.2.20200808..
Preprocessing library for ghc-lib-parser-8.10.2.20200808..
Building library for ghc-lib-parser-8.10.2.20200808..
[  1 of 228] Compiling GHC.BaseDir      ( libraries\ghc-boot\GHC\BaseDir.hs, dist\build\GHC\BaseDir.o )
....
Configuring library for ghc-lib-8.10.2.20200808..
Preprocessing library for ghc-lib-8.10.2.20200808..
Building library for ghc-lib-8.10.2.20200808..
....
[201 of 346] Compiling CmmParse         ( D:\ws\haskell\ghc-lib-8.10.2.20200808\dist-newstyle\build\x86_64-windows\ghc-8.8.4\ghc-lib-8.10.2.20200808\build\CmmParse.hs, D:\ws\haskell\ghc-lib-8.10.2.20200808\dist-newstyle\build\x86_64-windows\ghc-8.8.4\ghc-lib-8.10.2.20200808\build\CmmParse.o )

D:\\ws\haskell\ghc-lib-8.10.2.20200808\dist-newstyle\build\x86_64-windows\ghc-8.8.4\ghc-lib-8.10.2.20200808\build\CmmParse.hs:477:24: error:
    Not in scope: `Prelude.*'
    No module named `Prelude' is imported.
    |
477 |         bit_start = st Prelude.* 134
    |       
... lot of similar errors with other Prelude definitions
```

jneira:
Same for ghc-lib-8.10.1.20200523

shayne-fletcher:
Reproduced.

[...]

PR #244, opened 2020-09-16
On ghc-8.10.*, add missing import to CmmParse fixing ghc-lib build error

shayne-fletcher:
Fixes https://github.com/digital-asset/ghc-lib/issues/243. No idea why the issue doesn't manifest in CI; given though that the underlying resolution is already applied to master it doesn't seem worth spending more time looking at this.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -10,6 +10,7 @@ module Ghclibgen (
   , applyPatchRtsIncludePaths
   , applyPatchStage
   , applyPatchNoMonoLocalBinds
+  , applyPatchCmmParseNoImplicitPrelude
   , generatePrerequisites
   , mangleCSymbols
   , generateGhcLibCabal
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -550,6 +551,22 @@ applyPatchNoMonoLocalBinds _ =
...
+ GHC/cmm/Parser.y and already contains the missing directive so
+ nothing to do there. -}
+applyPatchCmmParseNoImplicitPrelude :: GhcFlavor -> IO ()
+applyPatchCmmParseNoImplicitPrelude ghcFlavor =
+  when (ghcFlavor == Ghc8101 || ghcFlavor == Ghc8102) $
+    writeFile "compiler/cmm/CmmParse.y" .
+      replace
+        "import GhcPrelude"
+        "import GhcPrelude\nimport qualified Prelude"
+    =<< readFile' "compiler/cmm/CmmParse.y"
```
hurryabit, line 568: Is `readFile'` a strict version of `readFile` that closes the file before returning anything? Otherwise, I'd be surprised if this works.

hurryabit (approved):
Looks good to me. Thank you.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Yes. From https://hoogle.haskell.org/?hoogle=readFile%27: "A strict version of readFile. When the string is produced, the entire file will have been read into memory and the file handle will have been closed. Closing the file handle does not rely on the garbage collector."

hurryabit: Thanks a lot for the explanation!

PR #245, opened 2020-09-20
Sync head

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git `057db94ce038970b14df1599fe83097c284b9c1f`
- Update the `stack.yaml` resolver to be consistent with the `hadrian/stack.yaml` resolver;
- Factor out `applyPatchHadrianStackYaml` and add a line for happy (https://gitlab.haskell.org/ghc/ghc/-/issues/18726);
- ~Needs rebasing on master after https://github.com/digital-asset/ghc-lib/commit/8c2dae9a2026b32b1d4782affd700e493fa3cff9 lands.~

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -56,7 +56,7 @@ data GhcFlavor = Ghc8101
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current = "8440b5fa1397940f2f293935927e690b34110a73" -- 2020-09-12
+current = "057db94ce038970b14df1599fe83097c284b9c1f" -- 2020-09-20
 
 -- Command line argument generators.
 
@@ -176,6 +176,33 @@ parseStackOptions = StackOptions
        <> Opts.help "If specified, pass '--ghc-options=\"xxx\"' to stack"
         ))
 
+-- [Note : GHC now depends on exceptions package]
+-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+-- As of
+-- https://gitlab.haskell.org/ghc/ghc/-/commit/30272412fa437ab8e7a8035db94a278e10513413
+-- (4th May 2020), certain GHC modules depend on the exceptions
+-- package. Depending on the boot compiler, this package may or may
+-- not be present and if it's missing, determining the list of
+-- ghc-lib-parser modules (via 'calcParserModules') will fail. The
+-- function 'appyPatchHadrianStackYaml' guarantees it is available
+-- if its needed.
+
+-- Patch Hadrian's Cabal.
+-- TODO (SF, 2020-09-20): Move this function and the above note to
+-- ghc-lib-gen.
+applyPatchHadrianStackYaml :: GhcFlavor -> IO ()
+applyPatchHadrianStackYaml ghcFlavor = do
+   appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
+   case ghcFlavor of
+     GhcMaster _ ->
+       appendFile "ghc/hadrian/stack.yaml" $
+       unlines [
+           "extra-deps:"
+         , "  - happy-1.20.0" -- See https://gitlab.haskell.org/ghc/ghc/-/issues/18726.
+         , "  - exceptions-0.10.4"
+         ]
+     _ -> pure ()
+
 buildDists :: GhcFlavor -> StackOptions -> IO String
 buildDists
   ghcFlavor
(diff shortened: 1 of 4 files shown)


CI.hs
```diff
@@ -176,6 +176,33 @@ parseStackOptions = StackOptions
...
+-- https://gitlab.haskell.org/ghc/ghc/-/commit/30272412fa437ab8e7a8035db94a278e10513413
+-- (4th May 2020), certain GHC modules depend on the exceptions
+-- package. Depending on the boot compiler, this package may or may
+-- not be present and if it's missing, determining the list of
+-- ghc-lib-parser modules (via 'calcParserModules') will fail. The
+-- function 'appyPatchHadrianStackYaml' guarantees it is available
+-- if its needed.
+
+-- Patch Hadrian's Cabal.
+-- TODO (SF, 2020-09-20): Move this function and the above note to
```
cocreature, line 191: Any reason not do this now? This seems pretty crucial. We don’t want to force people to use `CI`. `ghc-lib-gen` should work on its own.

shayne-fletcher: there's no deep reason just that i spotted in the course of this work it's in the wrong place. sure, let's knock it over. stand by.

shayne-fletcher:
When https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4096 containing https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4096/diffs?commit_id=e436b73f3687b7cbca3f2ab63d9d753660aabbce lands, we'll be able to unwind writing the extra dep.

CI.hs
shayne-fletcher: done

CI.hs
```diff
@@ -258,8 +244,6 @@ buildDists
 
     -- Make and extract an sdist of ghc-lib.
     cmd "cd ghc && git checkout ."
-    appendFile "ghc/hadrian/stack.yaml" $ unlines ["ghc-options:","  \"$everything\": -O0 -j"]
-    -- Feedback on the compiler used for ghc-lib-gen.
```
hurryabit, line 262: We didn't add any extra-deps here but we seem to add them globally with `applyPatchHadrianStackYaml` now. Is this intended?

shayne-fletcher: The extra-deps were added a little bit further down: https://github.com/digital-asset/ghc-lib/blob/6be486b61f07c022402d632e161ab081c837a74b/CI.hs#L247-L249. This block is now correctly moved to ghc-lib-gen and an additional extra dep added (which is temporary while the upstream fix isn't landed).

hurryabit: The line you're referring to is a bit further up. Let me rephrase what I find confusing: Before this change we had tw locations where we wrote `ghc-options` to `stack.yaml` and only location where we wrote `extra-deps`. Now, we have one location where we write both. What happened to the case where we need to write the `ghc-options` but not the `extra-deps`?

shayne-fletcher: Before in CI.hs, we unconditionally wrote ghc-options and wrote extra-deps conditionally on flavor ghc-master in two consecutive statements. These two statements have been moved to ghc-lib-gen and the interior of the function `applyPatchHadrianStackYaml`. The function will not write extra-deps if the flavor is not ghc-master just as it were before.

At the risk of confusing things further but in the interests of completeness, there is one further change that is being overlooked: CI.hs used to patch hadrian.cabal exactly once. This was a (then not,  but now important) error. It should happen twice because in-between generating ghc-lib-parser.cabal and ghc-lib.cabal, the changes to hadrian/hadrian.cabal are reverted. Now, `Main.hs` arranges to patch  hadrian's cabal  on every execution of ghc-lib-gen regardless of whether the target is ghc-lib-parser or ghc-lib.

shayne-fletcher: In any case, the key point is that the two statements that were in CI.hs are exactly next two each other in ghc-lib-gen now as they were before https://github.com/shayne-fletcher/ghc-lib/blob/03e3d68df6fcd132fe7a9b836c9cd90b0616cde0/ghc-lib-gen/src/Ghclibgen.hs#L584-L594.

shayne-fletcher: @hurryabit  Ah! I see what you are saying. Yes, we were only writing the extra-deps once before. That is what I am wittering on about here https://github.com/digital-asset/ghc-lib/pull/245#discussion_r492125958.  We are agreed that this was not correct and this PR incidentally fixes that.

hurryabit: That explains it. Thanks a lot.

hurryabit (approved):
Looks good to me. Thank you.

shayne-fletcher:
I've never seen anything like that Windows failure before. I suspect it will work if restarted.

cocreature:
We got the same windows error on the daml CI today and solved it by upgrading stack.

shayne-fletcher:
We can put a line in `azure-pipelines.yml`:
```
  - script: |
      curl -sSL https://get.haskellstack.org/ | sh
      stack upgrade
```
The effect then should be to guarantee stack is installed and that it is the most recent stack available. I'll push that and let's see.

shayne-fletcher:
I think we are actually looking at the systems of a stack outage. On my clone all builds are failing right now with the same issue. Re-running at this time probably won't help. I think we have to wait.

PR #246, opened 2020-09-23
sync head

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git `d7385f7077c6258c2a76ae51b4ea80f6fa9c7015`;
- Remove `happy-1.20.0` from extra-deps in hadrian stack.yaml now https://gitlab.haskell.org/ghc/ghc/-/commit/aaa51dcfdb729f130aeefeaeac15029b62096a74 has landed;
- Rework `applyPatchHadrianStackYaml` to be a little more resilient to the presence/absence of fields.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -35,6 +35,7 @@ import qualified Data.Set as Set
 
 import qualified Data.Text as T
 import qualified Data.Yaml as Y
+import qualified Data.Yaml.Pretty as Y
 import Data.Yaml (FromJSON(..), ToJSON(..), (.:?), (.!=))
 import qualified Data.ByteString as B
 import qualified Data.HashMap.Strict as HMS
@@ -590,7 +591,7 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
 
 data HadrianStackYaml =
   HadrianStackYaml {
-      extraDeps :: [T.Text]
+      extraDeps :: [Y.Value]
     , ghcOptions :: HMS.HashMap T.Text T.Text
     , otherFields :: Y.Object
   } deriving (Eq, Show)
@@ -620,8 +621,10 @@ applyPatchHadrianStackYaml ghcFlavor = do
  -- Parser.hs) as quickly as possible.
   let opts = HMS.insert "$everything" "-O0 -j" (ghcOptions config)
   -- See [Note : GHC now depends on exceptions package]
-  let deps = [x | ghcFlavor == GhcMaster, x <- ["exceptions-0.10.4"]] ++ extraDeps config
-  B.writeFile hadrianStackYaml $ Y.encode config{ extraDeps=deps, ghcOptions=opts }
+  let deps = [x | ghcFlavor == GhcMaster, x <- [toJSON (T.pack "exceptions-0.10.4")]] ++ extraDeps config
+  B.writeFile hadrianStackYaml $
+    Y.encodePretty (Y.setConfCompare compare (Y.setConfDropNull True Y.defConfig))
+    config{extraDeps=deps, ghcOptions=opts}
 
 -- | Data type representing an approximately parsed Cabal file.
 data Cabal = Cabal
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -3,7 +3,6 @@
 -- (Apache-2.0 OR BSD-3-Clause)
 
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
 
 module Ghclibgen (
     applyPatchHeapClosures
@@ -34,10 +33,9 @@ import Data.Ord
 import qualified Data.Set as Set
 
 import qualified Data.Text as T
+import Data.Aeson.Types(parse, Result(..))
 import qualified Data.Yaml as Y
-import qualified Data.Yaml.Pretty as Y
-import Data.Yaml (FromJSON(..), ToJSON(..), (.:?), (.!=))
-import qualified Data.ByteString as B
+import Data.Yaml (ToJSON(..), (.:?), (.!=))
 import qualified Data.HashMap.Strict as HMS
 
 import GhclibgenOpts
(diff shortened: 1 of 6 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +579,35 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
+   update fld elems =
+     writeFile hadrianStackYaml .
+       appendOrExtend fld elems
+       =<< readFile' hadrianStackYaml
+
+   -- If there are multiple occurences of a stanza (e.g. 'extra-deps')
+   -- they don't get merged and the last occurence prevails. Thus it's
+   -- important to extend a stanza if one already exists and not just
+   -- append a new one.
+   appendOrExtend :: String -> [String] -> String -> String
```
cocreature, line 604: This seems pretty messy. Can we just use `yaml`/`aeson` instead and parse this properly?

shayne-fletcher: That seems total overkill for this. Further, I don't think that stack.yaml is either yaml or json. Look at the format of ghc-options vs extra-deps for an example.

shayne-fletcher: well, that's wrong of course. presumably it's yaml.

cocreature: I’m not sure I’m following. Why are those fields not valid yaml? Parsing yaml into an Aeson object is hardly and modifying the field is hardly any longer than this, much more robust and easier to understand.

shayne-fletcher: yep. i'll see what i can do.

shayne-fletcher: thanks @cocreature i switched the code over to aeson/yaml as you suggested. let me know what you think!

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +589,42 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
 -- function 'appyPatchHadrianStackYaml' guarantees it is available
 -- if its needed.
 
--- Patch Hadrian's Cabal.
+data HadrianStackYaml =
+  HadrianStackYaml {
+      extraDeps :: [Y.Value]
+    , ghcOptions :: HMS.HashMap T.Text T.Text
+    , otherFields :: Y.Object
+  } deriving (Eq, Show)
```
hurryabit, line 597: Can't we just do the modifications on the YAML value without this type and its instances? That should be significantly shorter and probably not harder to understand.

shayne-fletcher: I'm sorry, you'll have to demonstrate the technique. I don't know how to do that.

shayne-fletcher: By the way, what would be the goal? Is there something wrong with this program?

cocreature: parse into `Object` and do a key lookup on the hashmap instead of parsing into a custom type.

shayne-fletcher: Ah right, I get you. Well, the more verbose version exists. Is there real value in unwinding it in favor of doing that?

shayne-fletcher: It is of course a short distance from here to there compared to from where we were.

shayne-fletcher: it's done

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +590,25 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
-       appendFile "hadrian/stack.yaml" $
-       unlines [
-           "extra-deps:"
-         , "  - happy-1.20.0" -- See https://gitlab.haskell.org/ghc/ghc/-/issues/18726.
-         , "  - exceptions-0.10.4" -- See [Note : GHC now depends on exceptions package]
-         ]
-     _ -> pure ()
+  let hadrianStackYaml = "hadrian/stack.yaml"
+  bytes <- B.readFile hadrianStackYaml
+  config <- Y.decodeThrow bytes
```
hurryabit, line 597:
```suggestion
  config <- Y.decodeFileThrow hadrianStackYaml
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +590,25 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
+  -- See [Note : GHC now depends on exceptions package]
+  let deps = [x | ghcFlavor == GhcMaster, x <- [T.pack "exceptions-0.10.4"]] ++
```
hurryabit:
```suggestion
  let deps = ["exceptions-0.10.4" | ghcFlavor == GhcMaster] ++
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +590,25 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
+        case parse (\cfg -> cfg .:? "extra-deps" .!= []) config of
+          Success ls -> ls :: [T.Text]
+          _ -> []
```
hurryabit, line 602: If `extra-deps` is not a list, something must have gone wrong. Let's not swallow the error but rather crash noisily.
```suggestion
          Error msg -> error msg
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +590,25 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
+ -- Build hadrian (and any artifacts we generate via hadrian e.g.
+ -- Parser.hs) as quickly as possible.
+  let opts = HMS.insert "$everything" "-O0 -j" $
+        case parse (\cfg -> cfg .:? "ghc-options" .!= HMS.empty) config of
+          Success os -> os :: HMS.HashMap T.Text Y.Value
+          _ -> HMS.empty
```
hurryabit, line 608:
```suggestion
          Error msg -> error msg
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -579,19 +589,25 @@ applyPatchCmmParseNoImplicitPrelude ghcFlavor =
...
-           "extra-deps:"
-         , "  - happy-1.20.0" -- See https://gitlab.haskell.org/ghc/ghc/-/issues/18726.
-         , "  - exceptions-0.10.4" -- See [Note : GHC now depends on exceptions package]
-         ]
-     _ -> pure ()
+  let hadrianStackYaml = "hadrian/stack.yaml"
+  bytes <- B.readFile hadrianStackYaml
+  config <- Y.decodeThrow bytes
+  -- See [Note : GHC now depends on exceptions package]
+  let deps = [x | ghcFlavor == GhcMaster, x <- [toJSON (T.pack "exceptions-0.10.4")]] ++
```
hurryabit, line 599: `Y.Value` has an `IsString` instance and we don't need the binding:
```suggestion
  let deps = ["exceptions-0.10.4" | ghcFlavor == GhcMaster] ++
```

hurryabit (approved):
Much better. Thank you.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Ok

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Ok

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Ok

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: ok

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Ok

shayne-fletcher:
Can we move this forward to `/azp run` please?

PR #248, opened 2020-09-29
Add support for ghc-9.0.1

shayne-fletcher:
This PR adds support for ghc-9.0.1 (currently in alpha).


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -223,22 +226,18 @@ compilerDependencies ghcFlavor =
     , "primop-vector-tys.hs-incl"
     , "primop-vector-uniques.hs-incl"
     ] ++
-    [ x | ghcFlavor == GhcMaster
-        , x <- [ "primop-docs.hs-incl"
-               , "GHC/Platform/Constants.hs"
-               ]
-    ]
+    [ "primop-docs.hs-incl" | ghcFlavor `elem` [ GhcMaster, Ghc901 ] ]
```
cocreature, line 229: What happened with `Constants.hs`?

cocreature (approved):
Nice! One question

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: It was being listed twice (and I judged was out of place in) `compilerDependencies`. It remains in `packageCode`.

PR #249, opened 2020-09-30
Fix CPP errors

shayne-fletcher:
PR https://github.com/digital-asset/ghc-lib/pull/248 added support for ghc-9.0.1 but messed up the CPP in the tests such that it was wrong but the tests succeeded anyway. This PR fixes the situation.


cocreature (approved):
good catch! Why did it succeed anyway?

shayne-fletcher:
> good catch! Why did it succeed anyway?

the old code allowed for `GHC_8101` to become defined and that enabled fluking successful tests.

Issue #256, opened 2020-10-16
Please normalise filepaths.


phadej:
The `ghc-lib-parser` has

```cabal
    hs-source-dirs;
        ...
        compiler/.
        compiler
```

The `Cabal-3.6` will warn on trailing dot paths. 

In fact, the `ghc-lib-parser` is *the only package* having `directory/.` i.e. trailing dot (not counting somewhat common `./.`) , otherwise I'd made parser fail out right on it.
(I still may make it so, have to think).

shayne-fletcher:
Thanks for pointing this out. Fix is on the way!

shayne-fletcher:
Fixed by https://github.com/digital-asset/ghc-lib/pull/257.

shayne-fletcher:
suggest close @dylant-da @garyverhaegen-da @akrmn @samuel-williams-da

PR #257, opened 2020-10-16
No trailing . in hs-source-dirs entries

shayne-fletcher:
Fixes issue https://github.com/digital-asset/ghc-lib/issues/256.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -22,8 +22,8 @@ module Ghclibgen (
 
 import Control.Monad
 import System.Process.Extra
-import System.FilePath hiding ((</>))
-import System.FilePath.Posix((</>)) -- Make sure we generate / on all platforms.
+import System.FilePath hiding ((</>), normalise, dropTrailingPathSeparator)
+import System.FilePath.Posix((</>), normalise, dropTrailingPathSeparator) -- Make sure we generate / on all platforms.
 import System.Directory.Extra
 import System.IO.Extra
 import Data.List.Extra hiding (find)
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -73,7 +73,7 @@ ghcLibParserIncludeDirs ghcFlavor =
 -- not the case).
 sortDiffListByLength :: Set.Set FilePath -> Set.Set FilePath -> [FilePath]
 sortDiffListByLength all excludes =
-  sortOn (Down . length) $ Set.toList (Set.difference all excludes)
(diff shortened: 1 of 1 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -100,7 +100,7 @@ ghcLibParserHsSrcDirs forParserDepends ghcFlavor lib =
         -- 'GhcVersion.hs' IIRC).
         [ stage0GhcBoot | ghcFlavor `elem` [ GhcMaster, Ghc8101, Ghc8102, Ghc901 ] ] ++
         map takeDirectory cabalFileLibraries ++
-        askFiles lib "hs-source-dirs:"
+        filter (not . isSuffixOf ".") (askFiles lib "hs-source-dirs:")
```
hurryabit, line 103: I'm not convinced this is the right approach. For instance, if we only have `foo/.` in the list but not `foo`, then there will be no `foo` left afterward. I think doing what was suggested in the issue and normalizing file paths seems more appropriate. [`System.FilePath.normalise`](https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:normalise) could be very helpful for this.

shayne-fletcher: I looked into this very deeply. I won't bother you with the details of why but the fix is correct (but am happy to do so if you want an explanation). That said I have no problems with throwing a `normalise` in the mix so stand by for that.

shayne-fletcher: I agree with you that using `normalise` is more robust and therefore better for future proofing.

shayne-fletcher:
Ready for another look when you are @hurryabit . Thank-you!

shayne-fletcher:
Before (this is on ghc-master; the trailing dot problem only entered recently after the API reorganization):
```
    hs-source-dirs:
        ghc-lib/stage0/libraries/ghc-boot/build
        ghc-lib/stage0/compiler/build
        libraries/template-haskell
        libraries/ghc-boot-th
        libraries/ghc-boot
        libraries/ghc-heap
        libraries/ghci
        compiler
        compiler/.
```
After:
```
    hs-source-dirs:
        ghc-lib/stage0/libraries/ghc-boot/build
        ghc-lib/stage0/compiler/build
        libraries/template-haskell
        libraries/ghc-boot-th
        libraries/ghc-boot
        libraries/ghc-heap
        libraries/ghci
        compiler
```

hurryabit (approved):
This looks much better. Thanks for fixing it!

shayne-fletcher:
I think the last commit will deal with the Windows build failure. May I have another `/azp run` please?

PR #261, opened 2020-11-29
Sync head; patch GHCi Message.hs

shayne-fletcher:
- Sync to git@gitlab.haskell.org:ghc/ghc.git @ `ae14f160c64d20880486ba365348ef3900c84a60`
- Synthesize definition of `MIN_VERSION_ghc_heap(8,11,0)` in `libraries/ghci/GHCi/Message.hs` when `--ghc-flavor=ghc-master` to account for [this ghc-heap MR](https://gitlab.haskell.org/ghc/ghc/-/commit/625726f988852f5779825a954609d187d9865dc1)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -408,6 +409,16 @@ applyPatchDisableCompileTimeOptimizations ghcFlavor =
           ("{-# OPTIONS_GHC -O0 #-}\n" ++)
           =<< readFile' file
 
+applyPatchGHCiMessage :: GhcFlavor -> IO ()
+applyPatchGHCiMessage ghcFlavor =
+  when (ghcFlavor == GhcMaster) $ do
+    writeFile messageHs .
+        replace
+          "#if MIN_VERSION_ghc_heap(8,11,0)"
+          "#ifndef MIN_VERSION_ghc_heap\n#define MIN_VERSION_ghc_heap(major1,major2,minor) ((major1) <  8 || (major1) == 8 && (major2) <  11 || (major1) == 8 && (major2) == 11 && (minor) <= 0)\n#endif /* MIN_VERSION_ghc_heap */\n#if MIN_VERSION_ghc_heap(8,11,0)"
```
hurryabit, line 418: Wouldn't it be better to do it as described in https://gitlab.haskell.org/ghc/ghc/-/blob/master/docs/users_guide/phases.rst#L429?

shayne-fletcher: @hurryabit Can you be more explicit please? Even after reading that link I don't know what you are recommending.

shayne-fletcher: Perhaps you mean, write it like this?
```
        replace
          "#if MIN_VERSION_ghc_heap(8,11,0)"
          "#if !defined(MIN_VERSION_ghc_heap) || MIN_VERSION_ghc_heap(8,11,0)"
```
If so, I agree it's better and am testing it now.

shayne-fletcher: Perhaps you mean, write it like this?
```
        replace
          "#if MIN_VERSION_ghc_heap(8,11,0)"
          "#if !defined(MIN_VERSION_ghc_heap) || MIN_VERSION_ghc_heap(8,11,0)"
```
If so, ~I agree it's better and am testing it now~ unfortunately, that doesn't work. Another approach would be to delete the guard. I'm not convinced that's in any way better than what we have here.

hurryabit: I meant replacing it with
```
#if defined(MIN_VERSION_ghc_heap)
#if MIN_VERSION_ghc_heap(8,11,0)
```
But I guess that's a bit cumbersome because we'd need two `#endif`s then. Replacing it with
```
#if defined(MIN_VERSION_ghc_heap) && MIN_VERSION_ghc_heap(8,11,0)
```
should do the job as well. The thing I'd like to avoid is redefining one of these `MIN_VERSION_xyz` macros.

Would that work?

shayne-fletcher: Normally, when building a program that depends on `ghc-heap`, you'd compile your program with `ghc -package ghc-heap ...`. GHC would then automatically produce a definition of `MIN_VERSION_ghc_heap`  for you that you can use in your program. In our case, we don't build against the `ghc-heap` package, rather we include its sources directly into our program. It follows that `MIN_VERSION_ghc_heap` is not generated and therefore not available to us.

Now, when we see

```
#if MIN_VERSION_ghc_heap(8,11,0)
instance Binary Heap.StgTSOProfInfo
instance Binary Heap.CostCentreStack
instance Binary Heap.CostCentre
instance Binary Heap.IndexTable
instance Binary Heap.WhatNext
instance Binary Heap.WhyBlocked
instance Binary Heap.TsoFlags
#endif
```

we want our program to behave as if `MIN_VERSION_ghc_heap(8,11,0)` holds and we get the guarded instance declarations. The way I've chosen to do that is (1) to synthesize the definition of `MIN_VERSION_ghc_heap` (after all, we are morally on that version we just don't have the macro definition) and inject it before the test. Another approach would be to (2) remove the guard and its corresponding `#endif` completely.

Your suggestion of 
```
#if defined(MIN_VERSION_ghc_heap)
#if MIN_VERSION_ghc_heap(8,11,0)
...
```
means that we won't get the definitions guarded by `MIN_VERSION_ghc_heap(8,11,0)` and the closely related alternative that the idea suggests  of saying `#if !defined(MIN_VERSION_ghc_heap) || MIN_VERSION_ghc_heap(8,11,0)` won't work because macros are fully expanded before they are evaluated so the bit after the `||` generates an error.

I can see other ways of getting the effect we want e.g. we can (3) generate this block (replicating the 8.11.0 case):

```
#if !defined(MIN_VERSION_ghc_heap)
instance Binary Heap.StgTSOProfInfo
instance Binary Heap.CostCentreStack
instance Binary Heap.CostCentre
instance Binary Heap.IndexTable
instance Binary Heap.WhatNext
instance Binary Heap.WhyBlocked
instance Binary Heap.TsoFlags
#endif
```

You might prefer this as it avoids us generating a (pseudo) definition for `MIN_VERSION_ghc_heap`? In any case I think our choices are basically one of (1), (2) or (3).

hurryabit: Thanks for the detailed explanation. I wasn't aware that `MIN_VERSION_ghc_heap` is different from `MIN_VERSION_GLASGOW_HASKELL`.

Doesn't what the PR is doing so far amount to hardcoding the version of `ghc-heap` to 8.11.0? Shouldn't we instead replace with
```
#ifndef MIN_VERSION_ghc_heap
#define MIN_VERSION_ghc_heap(major1,major2,minor) ((major1) <  X || (major1) == X && (major2) <  Y || (major1) == X && (major2) == Y && (minor) <= Z)
#endif /* MIN_VERSION_ghc_heap */
#if MIN_VERSION_ghc_heap(8,11,0)
```
where `X.Y.Z` is the version of `ghc-heap` whose files we're copying? Can we actually get our hands on that version number? (Wrapping my head around these `MIN_VERSION_...` macros has just made it explode... 🤯)

shayne-fletcher: > Doesn't what the PR is doing so far amount to hardcoding the version of ghc-heap to 8.11.0?

Yes.

> Shouldn't we instead replace with
> [...]
> where X.Y.Z is the version of ghc-heap whose files we're copying?

That would be an improvement.

> Can we actually get our hands on that version number? 

Yes. I've reached the conclusion that`ghc-heap` (`ghc-boot`, `ghc-boot-th`, `ghci`) version numbers are coincident with the GHC version number determined by `./configure` and the contents of `configure.ac` (makes sense in retrospect). Accordingly, we can get the prevailing `X.Y.Z` by inspection of the contents of the `ghcversion.h` we generate. I'll push an update to this PR now that does that.

shayne-fletcher:
@hurryabit trouble you for another look sir?

gettyimages-3205380-612x612

ghc-lib-gen/src/Ghclibgen.hs
hurryabit: If we work under the assumption that `ghc-heap` has the same version as GHC, can't we just replace `MIN_VERSION_ghc_heap` by `MIN_VERSION_GLASGOW_HASKELL` then?

shayne-fletcher: @hurryabit  I expect you mean `__GLASGOW_HASKELL__` and writing `MIN_VERSION_ghc_heap` in terms of that instead of reading `__GLASGOW_HASKELL__` from `ghc_version.h` as my patch currently does?

If so, no because the prevailing version of `__GLASGOW_HASKELL__` is that of the compiler being used to build ghc-lib, not the version of GHC sources being built.

hurryabit: That's a very good point. Thanks for catching my bad reasoning.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -408,6 +410,46 @@ applyPatchDisableCompileTimeOptimizations ghcFlavor =
...
+    -- Synthesize a definition of MIN_VERSION_ghc_heap. If X.Y.Z is
+    -- the current version, then MIN_VERSION_ghc_heap(a, b, c) is a
+    -- test of whether X.Y.Z >= a.b.c (that is, ghc-heap X.Y.Z is at
+    -- least a.b.c).
+    versionText <- readFile' (root </> "ghcversion.h")
+    let ls = lines versionText
+        t1 = "#define __GLASGOW_HASKELL__ "
+        t2 = "#define __GLASGOW_HASKELL_PATCHLEVEL1__ "
+        version      = dropPrefix t1 (head (filter (t1 `isPrefixOf`) ls))
+        patchLevel   = dropPrefix t2 (head (filter (t2 `isPrefixOf`) ls))
```
hurryabit, line 425:
```suggestion
        Just version <- firstJust (stripPrefix "#define __GLASGOW_HASKELL__ ") ls
        Just patchLevel <- firstJust (stripPrefix "#define __GLASGOW_HASKELL_PATCHLEVEL1__ ") ks
```

hurryabit (approved):
I'm happy with what we have now. Sorry for being such a pain in the butt but I think we've improved things along the way. Thank you such much, also for your patience.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Oh! That's very nice 🤩

shayne-fletcher:
> I'm happy with what we have now. Sorry for being such a pain in the butt but I think we've improved things along the way. Thank you such much, also for your patience.

Great teamwork!

shayne-fletcher:
@hurryabit Yay! Much greeness. Landing please 🙏🏻

Southern-Flying-Squirrel-Photo-Credit-Joe-McDonald2

shayne-fletcher:
Thank-you @cocreature 😉

PR #262, opened 2020-12-12
Sync head

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git `c647763954717d9853d08ff04eece7f1ddeae15c`
- Patch `removeGeneratedIntermediateFiles` to accommodate [this commit](https://gitlab.haskell.org/ghc/ghc/-/commit/3a16d764f3cf01add8c09b9ca5c071176f857fb8)

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -735,15 +735,16 @@ libBinParserModules ghcFlavor = do
 -- the sdists rather than these. If we don't remove them, some
 -- ambiguity results. Cabal and stack are OK with that but bazel gets
 -- messed up.
-removeGeneratedIntermediateFiles :: IO ()
-removeGeneratedIntermediateFiles = do
+removeGeneratedIntermediateFiles :: GhcFlavor -> IO ()
+removeGeneratedIntermediateFiles ghcFlavor = do
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/Utils.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/InfoTableProf.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/InfoTable.hs"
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -735,15 +735,16 @@ libBinParserModules ghcFlavor = do
...
+removeGeneratedIntermediateFiles ghcFlavor = do
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/Utils.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/InfoTableProf.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/InfoTable.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/InfoTable/Types.hs"
     removeFile $ stage0GhcHeap </> "GHC/Exts/Heap/Constants.hs"
     removeFile $ stage0Ghci </> "GHCi/FFI.hs"
-    removeFile $ stage0Ghci </> "GHCi/InfoTable.hs"
+    when (ghcFlavor /= GhcMaster) $
+      removeFile $ stage0Ghci </> "GHCi/InfoTable.hs"
```
hurryabit, line 747: Why do we only remove `GHCi/InfoTable.hs` but not `GHCi/StaticPtrTable.hs`, which seems to have changed in the same way?

shayne-fletcher: It is not a "generated intermediate file" as `InfoTable.hs` is.

hurryabit: I see. Thanks for the explanation.

hurryabit (approved):
This looks good to me. Thanks.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: You bet.

shayne-fletcher:
> This looks good to me. Thanks.

Thanks @hurryabit !

PR #264, opened 2020-12-20
Add support for ghc-8.10.3

shayne-fletcher:
Add support for ghc-8.10.3:
- Allow builds of `ghc-lib` from the `ghc-8.10.3-released` tag;
- Have Azure test `ghc-flavor=ghc-8.10.3`  now that ghc-8.10.2 is eclipsed;
- Tiny but long overdue refactoring of the Hadrian generated root path calculation to reduce code repetition.

Enable ghc-8.10.3 as build compiler:
- Update `stack-exact.yaml` so that
  - Default resolver is `ghc-8.10.3`;
  - `extra-deps` updated in line with the what ships with that resolver  and the most recent packages now available;
  - Refine `applyPatchCmmParseNoImplicitPrelude` (previous version was found to depend not so much on `ghc-flavor` as the build compiler; this addresses the observed shortcoming).
  
Sync head:
- Sync to `26a928b8fdb1b4ccb75e8edb620b8cf12cb38621` @ https://gitlab.haskell.org/ghc/ghc.git.


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -100,7 +94,7 @@ ghcLibParserHsSrcDirs forParserDepends ghcFlavor lib =
         -- This next conditional just smooths over a 'master'
         -- vs. '8.8.1' branch difference (relating to a file
         -- 'GhcVersion.hs' IIRC).
-        [ stage0GhcBoot | ghcFlavor `elem` [ GhcMaster, Ghc8101, Ghc8102, Ghc901 ] ] ++
+        [ stage0GhcBoot | ghcFlavor `elem` [ GhcMaster, Ghc8101, Ghc8102, Ghc8103, Ghc901 ] ] ++
```
hurryabit, line 97: This list seems to get longer and longer. Would it make sense to put an order on the `GhcXYZ`s and replace this with something like
```suggestion
        [ stage0GhcBoot | ghcFlavor >= Ghc8101 ] ++
```
Same question in multiple places below.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -609,15 +592,22 @@ applyPatchNoMonoLocalBinds _ =
  https://github.com/digital-asset/ghc-lib/issues/243 for where the
  problem was first reported. On master, this file has moved to
  GHC/cmm/Parser.y and already contains the missing directive so
- nothing to do there. -}
+ nothing to do there.
+
+Update 30th Dec, 2020: This appears to be more to do with the compiler
+in play than the flavor. It's still mysterious but not interseting
```
hurryabit, line 598:
```suggestion
in play than the flavor. It's still mysterious but not interesting
```

hurryabit (approved):
Although I'm generally speaking not a big fan of doing too many things in a single PR, particularly not changing code and CI at the same time when not necessary, this looks good to me. Thank you so much for taking care of this.

And: Happy New Year, @shayne-fletcher!

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Ok, https://github.com/digital-asset/ghc-lib/issues/265.

shayne-fletcher:
> And: Happy New Year, @shayne-fletcher!

Happy New Year @hurryabit !

shayne-fletcher:
@hurryabit Landing please 🙏

Issue #271, opened 2021-02-06
ghc-lib-parser doesn't compile on ghc 9.0.1


GeorgeCo:

shayne-fletcher:
Version 9.0.1.20210207 is now on Hackage. Could you give that a try please?

GeorgeCo:
It installs now. Thanks, that was very quick!

PR #272, opened 2021-02-07
Enable ghc-9.0 as a build compiler

shayne-fletcher:
- Updates: `base < 4.16`, `ghc-prim < 0.8`;
- The logic for `applyPatchGhcPrim` needs an update now there is the possibility of `ghc-prim-0.7` (which ships with ghc-9.0.1);
- Fixes https://github.com/digital-asset/ghc-lib/issues/271.

I've hand verified these fixes using Cabal. Unfortunately building with `stack` and resolver `ghc-9.0` is broken preventing me from adding tests in this PR.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -446,10 +446,10 @@ applyPatchGhcPrim ghcFlavor = do
       writeFile tysPrim .
           replace
               "bcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep"
-              "\n#if __GLASGOW_HASKELL >= 900\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep\n#else\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName UnliftedRep\n#endif\n" .
+              "\n#if __GLASGOW_HASKELL__ >= 900\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep\n#else\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName UnliftedRep\n#endif\n" .
           replace
               "bcoPrimTyConName              = mkPrimTc (fsLit \"BCO\") bcoPrimTyConKey bcoPrimTyCon"
-              "\n#if __GLASGOW_HASKELL >= 900\nbcoPrimTyConName              = mkPrimTc (fsLit \"BCO\") bcoPrimTyConKey bcoPrimTyCon\n#else\nbcoPrimTyConName              = mkPrimTc (fsLit \"BCO#\") bcoPrimTyConKey bcoPrimTyCon\n#endif\n"
+              "\n#if __GLASGOW_HASKELL__ >= 900\nbcoPrimTyConName              = mkPrimTc (fsLit \"BCO\") bcoPrimTyConKey bcoPrimTyCon\n#else\nbcoPrimTyConName              = mkPrimTc (fsLit \"BCO#\") bcoPrimTyConKey bcoPrimTyCon\n#endif\n"
         =<< readFile' tysPrim
         )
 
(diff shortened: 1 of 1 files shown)


shayne-fletcher:
[This issue](https://github.com/commercialhaskell/stack/issues/5486) tracks the problem building packages with stack and ghc-9.0.1.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -446,33 +446,37 @@ applyPatchGhcPrim ghcFlavor = do
       writeFile tysPrim .
           replace
               "bcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep"
-              "bcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName UnliftedRep" .
+              "\n#if __GLASGOW_HASKELL >= 900\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep\n#else\nbcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName UnliftedRep\n#endif\n" .
```
cocreature, line 449: maybe throw some `unlines` in there and spread the text over multiple lines to make it easier to read?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -446,33 +446,37 @@ applyPatchGhcPrim ghcFlavor = do
...
     let createBCO = "libraries/ghci/GHCi/CreateBCO.hs"
     when (ghcFlavor >= Ghc901) (
       writeFile createBCO .
+          replace
+              "{-# LANGUAGE RecordWildCards #-}"
+              "{-# LANGUAGE RecordWildCards #-}\n{-# LANGUAGE CPP #-}" .
           replace
               "do linked_bco <- linkBCO' arr bco"
-              "do BCO bco# <- linkBCO' arr bco" .
+              "\n#if __GLASGOW_HASKELL__ >= 900\n     do linked_bco <- linkBCO' arr bco\n#else\n     do BCO bco# <- linkBCO' arr bco\n#endif" .
```
cocreature, line 464: shouldn’t this check on ghc-prim instead of GHC? Given how closely they are tied together, it probably doesn’t matter too much but it is a bit confusing.

cocreature, line 464: These all follow the same pattern. Match on something, wrap it in the if and don’t modify it for GHC >= 9.0, do something else for older GHCs. Can we factor this out?

ghc-lib-gen/src/Ghclibgen.hs
cocreature: Or even better as suggested below, put it in a helper function.

cocreature (approved):
LGTM with a few comments, thanks!

shayne-fletcher:
@cocreature good comments. if you don't mind, let's capture these in issues, land this and address in a follow up

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Issue https://github.com/digital-asset/ghc-lib/issues/275

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Issue https://github.com/digital-asset/ghc-lib/issues/275

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Issue https://github.com/digital-asset/ghc-lib/issues/275

shayne-fletcher:
> @cocreature good comments. if you don't mind, let's capture these in issues, land this and address in a follow up

https://github.com/digital-asset/ghc-lib/issues/275

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: i think we'll want `if MIN_VERSION_ghc_prim(0, 7, 0)` in the meantime, the line you comment indicates an error. should be `__GLASGOW_HASKELL__`. i'll push a fix for that.

cocreature:
Let’s please at least fix the `ghc-prim` stuff in this PR.

shayne-fletcher:
We can do that. Give me a few hours.

PR #273, opened 2021-02-10
ghc-in-ghci: Drop it

shayne-fletcher:
In [this discussion](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4968#note_329341) it is noted that `GHC_IN_GHCI` is no longer in use.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -778,7 +778,7 @@ generateGhcLibCabal ghcFlavor = do
         indent2 (ghcLibIncludeDirs ghcFlavor) ++
         [ "    ghc-options: -fobject-code -package=ghc-boot-th -optc-DTHREADED_RTS"
         , "    cc-options: -DTHREADED_RTS"
-        , "    cpp-options: " <> ghcStageDef ghcFlavor <> " -DTHREADED_RTS " <> ghciDef ghcFlavor <> " -DGHC_IN_GHCI"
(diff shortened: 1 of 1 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -804,22 +804,13 @@ generateGhcLibCabal ghcFlavor = do
...
-ghcStageDef GhcMaster = ""
-ghcStageDef Ghc901 = ""
-ghcStageDef Ghc8101 = ""
-ghcStageDef Ghc8102 = ""
-ghcStageDef Ghc8103 = ""
-ghcStageDef Ghc8104 = ""
-ghcStageDef _ = "-DSTAGE=2"
+ghcStageDef ghcFlavor = if ghcFlavor >= Ghc8101 then "" else "-DSTAGE=2"
+
+ghcInGhciDef :: GhcFlavor -> String
```
cocreature, line 812: let’s leave it out of this PR but returning empty strings seems a bit of an odd choice. We should return a `Maybe String` or a `[String]` and then throw `unwords` at it at the call site.

cocreature (approved):
Thanks!

shayne-fletcher:
@cocreature yep, let's capture this comment an issue, land this and do a refactoring in a follow up. thanks for the review and feedback!!

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: Issue: https://github.com/digital-asset/ghc-lib/issues/274

PR #277, opened 2021-02-17
Fix broken mini-compile test

shayne-fletcher:
- Failed to notice but mini-compile test broke recently, this fixes it
- Also take the chance to sync to https://gitlab.haskell.org/ghc/ghc.git @ `dbf8f6fe12db67b96412015a01646ce800f9988a`

### DIFF
--- a/examples/mini-compile/test/MiniCompileTest.hs
+++ b/examples/mini-compile/test/MiniCompileTest.hs
@@ -82,6 +82,7 @@ data KindRep = KindRepTyConApp TyCon [KindRep]
 
 data TypeLitSort = TypeLitSymbol
                  | TypeLitNat
+                 | TypeLitChar
 
 data TyCon = TyCon Word# Word#           -- Fingerprint
                    Module                -- Module in which this is defined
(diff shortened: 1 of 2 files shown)


examples/mini-compile/test/MiniCompileTest.hs
```diff
@@ -82,6 +82,7 @@ data KindRep = KindRepTyConApp TyCon [KindRep]
 
 data TypeLitSort = TypeLitSymbol
                  | TypeLitNat
+                 | TypeLitChar
```
cocreature, line 85: maybe worth thinking about testing this on CI so we notice it breaks next time.

cocreature (approved):
Thanks!

examples/mini-compile/test/MiniCompileTest.hs
shayne-fletcher: yeah. we need an improvement. expect testing would have caught this. i'll raise an issue.

shayne-fletcher: https://github.com/digital-asset/ghc-lib/issues/278

PR #279, opened 2021-02-28
Sync head

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git @ `915daf51357175fcc31a37c0aaf2347875560269`;
- Don't put the generated `ghcversion.h` in the dist.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -765,7 +765,7 @@ generateGhcLibCabal ghcFlavor = do
         , "bug-reports: https://github.com/digital-asset/ghc-lib/issues"
         , "data-dir: " ++ dataDir
         , "data-files:"] ++ indent dataFiles ++
-        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor) ++
+        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor \\ [hadrianGeneratedRoot ghcFlavor </> "ghcversion.h"]) ++
         [ "source-repository head"
         , "    type: git"
         , "    location: git@github.com:digital-asset/ghc-lib.git"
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -765,7 +765,7 @@ generateGhcLibCabal ghcFlavor = do
         , "bug-reports: https://github.com/digital-asset/ghc-lib/issues"
         , "data-dir: " ++ dataDir
         , "data-files:"] ++ indent dataFiles ++
-        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor) ++
+        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor \\ [hadrianGeneratedRoot ghcFlavor </> "ghcversion.h"]) ++
```
cocreature, line 768: why do we need to exclude this file? `ghcversion.h` has existed for a long time hasn’t it? did it change place?

shayne-fletcher: no, we've just been fortunate before now. what i determined is, there is a CMM .x file (lexer file). it now contains logic conditional on __GLASGOW_HASKELL__. if you build ghc-lib with a non ghc-9.0 via cabal (not stack), it's clear that in this configuration ghcversion.h gets in the way of the compiler provided ghc version.  so in summary, we need ghcversion.h to produce the dist but when building the dist, when you think about it, there is no scenario when substituting for __GLASGOW_HASKELL__ is the right thing to do - ghcversion.h should not be shipped.

shayne-fletcher: btw, in case you were wondering, ghcversion.h literally only contains a definition of `__GLASGOW_HASKELL__` and related constants. nothing else.

cocreature: makes sense, thanks for the explanation!

cocreature (approved):
Thanks!

shayne-fletcher:
> Thanks!

thank-you @cocreature for your continued support! it's much appreciated ❤️

Issue #275, opened 2021-02-14
Refactor `applyPatchGhcPrim`


shayne-fletcher:
Current implementation could do with refactoring for readability. Address the comments raised in https://github.com/digital-asset/ghc-lib/pull/272

shayne-fletcher:
- Confirmed the substitution of `#if __GLASGOW_HASKELL__ >= 900` with `#if MIN_VERSION_ghc_prim(0, 7, 0)` is valid with `ghc-9.0.1` as the build compiler

PR #280, opened 2021-03-08
A refactor of applyPatchGhcPrim

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git @ `3e082f8ff5ea2f42c5e6430094683b26b5818fb8`;
- In PR https://github.com/digital-asset/ghc-lib/pull/272 @cocreature [suggests a refactoring of `applyPatchGhcPrim`](https://github.com/digital-asset/ghc-lib/pull/272#discussion_r575858054). This does as suggested and fixes https://github.com/digital-asset/ghc-lib/issues/275.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -444,12 +444,12 @@ applyPatchGhcPrim ghcFlavor = do
             else "prelude/TysPrim.hs"
     when (ghcFlavor >= Ghc901) (
       writeFile tysPrim .
-          replace
-              "bcoPrimTyCon = pcPrimTyCon0 bcoPrimTyConName LiftedRep"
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -459,26 +459,34 @@ applyPatchGhcPrim ghcFlavor = do
...
+            "newBCO# instrs lits ptrs arity bitmap s"
+            "case newBCO# instrs lits ptrs arity bitmap s of\n    (# s1, bco #) -> (# s1, BCO bco #)"
         =<< readFile' createBCO
         )
+  where
+    replaceIfGhcPrim070Else :: Int -> String -> String -> String -> String
+    replaceIfGhcPrim070Else n s r = replace s (ifGhcPrim070Else n s r)
+
+    ifGhcPrim070Else :: Int -> String -> String -> String
+    ifGhcPrim070Else n s r =
```
cocreature, line 487: Relying on the caller passing in the indentation seems a bit fragile. We could instead see how much indentation there already is and just use that.

cocreature (approved):
Thanks!

Issue #282, opened 2021-03-15
Build failure with GHC-8.6


phadej:
```
[ 80 of 264] Compiling GHC.Types.Unique.Supply ( compiler/GHC/Types/Unique/Supply.hs, dist/build/GHC/Types/Unique/Supply.o )

compiler/GHC/Types/Unique/Supply.hs:299:10: error:
    Not in scope: type constructor or class ‘MonadFail’
    Perhaps you meant ‘MonadFix’ (imported from GHC.Utils.Monad)
    |
299 | instance MonadFail UniqSM where
    |          ^^^^^^^^^
cabal: Failed to build ghc-lib-parser-9.0.1.20210207 (which is required by
exe:hlint from hlint-3.3). See the build log above for details.
```

missing lower bounds?

See https://matrix.hackage.haskell.org/#/package/ghc-lib-parser

shayne-fletcher:
Thanks for the heads-up. No, expected failure. To bootstrap ghc-9.0.1 one needs a build compiler >= 8.8.1 (since 2020/03/31, https://gitlab.haskell.org/ghc/ghc/-/commit/57b888c0e90be7189285a6b078c30b26d092380). To bootstrap HEAD one needs >= 8.10.1 (since 2021/03/10, https://gitlab.haskell.org/ghc/ghc/-/commit/0a709dd9876e40c19c934692415c437ac434318c).

Any suggestions on how to stop this being red on the matrix builder? Maybe it's a matter of taking 8.6 out of the "tested at" section of the Cabal files?

phadej:
Use `base` lower bounds as proxy for GHC versions. In fact, it's not even GHC version, because that failure is really `base` library issue, not a some language feature present only in GHC-8.8+ (for which you should use `other-extensions: ...` and double check with `base` lower bound).

hamishmack:
Please use hackage `edit package information` feature to fix the lower bound on the following packages (should be `base >=4.13`):

https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210205
https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210207

shayne-fletcher:
> Please use hackage `edit package information` feature to fix the lower bound on the following packages (should be `base >=4.13`):
> 
> https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210205
> https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210207

Done

[...]

PR #284, opened 2021-03-15
Better base bounds

shayne-fletcher:
For each `ghc-lib` "flavor" (e.g. `8.8.*`, `8.10.*`, `9.0*`, ...) there is a minimum version requirement on the ghc build compiler. For example, for `ghc-lib` 8.8 parse trees, the minimum version is 8.4.4, for `ghc-lib` 8.10 parse, trees the minimum version is 8.8.1 and so on like this.

As evidenced on the [Hackage Matrix builder page](https://matrix.hackage.haskell.org/#/package/ghc-lib-parser) page, it would be nice to make Cabal build plans impossible if the minimum required build compiler version isn't met for a given build. See https://github.com/digital-asset/ghc-lib/issues/282#issuecomment-799749806 where @phadej (Oleg Genrus) suggests using base as a proxy for the minimum version. This PR implements that idea.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,13 +97,13 @@ strategy:
     # +---------+-----------------+------------+
     # | OS      | ghc-lib flavour | GHC        |
     # +=========+=================+============+
-    # | linux   | ghc-master      | ghc-8.8.4  |
+    # | linux   | ghc-master      | ghc-8.10.4 |
     # +---------+-----------------+------------+
-    linux-ghc-master-8.8.4:
+    linux-ghc-master-8.10.4:
       image: "ubuntu-latest"
       mode: "--ghc-flavor ghc-master"
       stack-yaml: "stack.yaml"
-      resolver: "nightly-2020-03-14"
+      resolver: "nightly-2021-03-07"
 
     # +---------+-----------------+------------+
     # | OS      | ghc-lib flavour | GHC        |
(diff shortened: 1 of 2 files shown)


phadej:
You technically have upper bounds too. It's accidental that e.g. 8.8 code compiles with GHC-9.0. Or do you actually do something for that. I don't think there is any guarantees that older GHC are bootable with newer compilers.

shayne-fletcher:
> You technically have upper bounds too. It's accidental that e.g. 8.8 code compiles with GHC-9.0. Or do you actually do something for that. I don't think there is any guarantees that older GHC are bootable with newer compilers.

I've no specific intent. My attitude is, if you try that and it happens to work great, if it doesn't, then it doesn't. If we don't have a specific motivation (like the matrix builder being red for example,) to tighten the upper bounds, my inclination is to let it be.

phadej:
Sorry, i missed that you do have upper bounds. It was only lower bounds which bit-rotted. Everything is fine then, as long as you test that `ghc-lib-parser-8.8` is in fact buildable with e.g. ghc-9.0.

azure-pipelines.yml
```diff
@@ -116,17 +116,6 @@ strategy:
...
-    # | OS      | ghc-lib flavour | GHC        |
-    # +=========+=================+============+
-    # | linux   | ghc-8.10.3      | ghc-8.8.4  |
-    # +---------+-----------------+------------+
-    linux-ghc-8.10.3-8.8.4:
-      image: "ubuntu-latest"
-      mode: "--ghc-flavor ghc-8.10.3"
-      resolver: "nightly-2020-03-14"
-      stack-yaml: "stack.yaml"
-
```
cocreature, line 129: Why drop this? Not opposed to it, I just don’t quite get the motivation.

cocreature (approved):
Thanks!

PR #285, opened 2021-03-16
Restore test

shayne-fletcher:
- Restore test (`ghc-flavor=8.10.3` with ghc-8.8.4)
  - As mentioned in https://github.com/digital-asset/ghc-lib/pull/284#discussion_r594917519
  -  Removed mistakenly
 - Tighten upper bounds on base ever so slightly.


cocreature (approved):

shayne-fletcher:
It turns out that tightening the upper bounds as I did wasn't a great idea. It breaks builds known to work like ghc-flavor 8.8 with ghc-8.10. Backing that out.

PR #287, opened 2021-03-20
Sync head: rts bytecode & in-tree annotations

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git @ `3e082f8ff5ea2f42c5e6430094683b26b5818fb8`;
- MR [Generate GHCi bytecode from STG instead of Core and support unboxed tuples and sums](https://gitlab.haskell.org/ghc/ghc/-/commit/1f94e0f7601f8e22fdd81a47f130650265a44196):
  - New instruction codes are added to `includes/rts/Bytecodes.h` and new Haskell code added that uses them:
  - When building`ghc-lib` we use the build compiler RTS:
    - Patch `GHC.ByteCode.Asm` to define these codes if the build compiler is <= 9.0.1;
    - Sufficient to keep `ghc-lib` compiling and its tests passing for now;
- Annotations code affected by MR [GHC Exactprint main commit](https://gitlab.haskell.org/ghc/ghc/-/commit/95275a5f25a2e70b71240d4756109180486af1b1):
  - Adjust the code to stop harvesting any annotations for the time being.

### DIFF
--- a/examples/mini-hlint/src/Main.hs
+++ b/examples/mini-hlint/src/Main.hs
@@ -207,13 +207,18 @@ analyzeExpr flags (L loc expr) =
                       ++ "`" ++ showSDoc flags (ppr expr) ++ "'")
     _ -> return ()
 
-#if defined (GHC_MASTER) || defined (GHC_901)
+#if defined (GHC_MASTER)
+analyzeModule :: DynFlags -> Located HsModule -> IO ()
+#elif defined (GHC_901)
 analyzeModule :: DynFlags -> Located HsModule -> ApiAnns -> IO ()
 #else
 analyzeModule :: DynFlags -> Located (HsModule GhcPs) -> ApiAnns -> IO ()
 #endif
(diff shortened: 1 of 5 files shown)


examples/mini-hlint/src/Main.hs
```diff
@@ -259,7 +267,9 @@ main = do
 #endif
         ]
     harvestAnns pst =
```
cocreature, line 269: Can we just not define this instead of setting it to `undefined`? Afaict, the callsite is guarded behind `!defined GHC_MASTER` so I don’t see why we need this there.

shayne-fletcher: Thanks, yes that's better. Just building here and then I'll push an amendment.

cocreature (approved):
Thanks!

shayne-fletcher:
> Thanks!

Thank-you @cocreature .

PR #289, opened 2021-03-31
sync head; ghc-lib depends on rts; fix autoconf (was Sync head; applyPatchGHCiInfoTable)

shayne-fletcher:
- Sync to https://gitlab.haskell.org/ghc/ghc.git @ `efe5fdab01012fae9436f5f8a9c67170ff185243`;
- ~~Write a patch for `libraries/ghci/GHCi/InfoTable.hsc`~~
  - Require a definition for `MIN_VERSION_rts` due to [this commit](https://gitlab.haskell.org/ghc/ghc/-/commit/e754ff7f178a629a2261cba77a29d9510391aebd);
- Small tweak to `strip-locs` for in-tree exact-print annotations;
- ~~Patch azure on macOS to select autoconf 2.69 over 2.71 which brew is now installing and doesn't work.~~

See https://github.com/digital-asset/ghc-lib/pull/289#issuecomment-812943544 for updated details on what's in this PR.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -156,8 +156,11 @@ steps:
   # macOS
   - bash: |
       /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-      brew install automake
+      brew install autoconf automake pkgconfig m4 libtool gettext
       brew upgrade gmp
+      curl -O -L http://ftpmirror.gnu.org/autoconf/autoconf-2.69.tar.gz
+      tar -xzf autoconf-2.69.tar.gz
+      (cd autoconf-* && ./configure && make && make install && autoconf --version)
     condition: eq( variables['Agent.OS'], 'Darwin' )
     displayName: Install brew
   - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -155,12 +155,8 @@ pool: {vmImage: '$(image)'}
 steps:
   # macOS
   - bash: |
-      /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-      brew install autoconf automake pkgconfig m4 libtool gettext
-      brew upgrade gmp
-      curl -O -L http://ftpmirror.gnu.org/autoconf/autoconf-2.69.tar.gz
-      tar -xzf autoconf-2.69.tar.gz
-      (cd autoconf-* && ./configure && make && make install && autoconf --version)
+      /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+      brew install automake
     condition: eq( variables['Agent.OS'], 'Darwin' )
     displayName: Install brew
   - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -156,11 +156,8 @@ steps:
   # macOS
   - bash: |
       /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-      brew install autoconf automake pkgconfig m4 libtool gettext
+      brew install autoconf@2.13 automake pkgconfig m4 libtool gettext
       brew upgrade gmp
-      curl -O -L http://ftpmirror.gnu.org/autoconf/autoconf-2.69.tar.gz
-      tar -xzf autoconf-2.69.tar.gz
-      (cd autoconf-* && ./configure && make && make install && autoconf --version)
     condition: eq( variables['Agent.OS'], 'Darwin' )
     displayName: Install brew
   - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -156,8 +156,11 @@ steps:
   # macOS
   - bash: |
       /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-      brew install autoconf@2.13 automake pkgconfig m4 libtool gettext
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -156,8 +156,11 @@ steps:
   # macOS
   - bash: |
       /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-      brew install automake
+      brew install autoconf automake pkgconfig m4 libtool gettext
```
cocreature, line 159: Why do we stillneed to install autoconf? To bootstrap the older version? More generally, it looks like brew has at least `autoconf@2.13`. Does that work? That seems like a nicer solution than building our own.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -404,6 +405,41 @@ applyPatchDisableCompileTimeOptimizations ghcFlavor =
           ("{-# OPTIONS_GHC -O0 #-}\n" ++)
           =<< readFile' file
 
+applyPatchGHCiInfoTable :: GhcFlavor -> IO ()
+applyPatchGHCiInfoTable ghcFlavor =
+  -- On master, we now require a definition of MIN_VERSION_RTS (since
```
cocreature, line 410:
```suggestion
  -- On master, we now require a definition of MIN_VERSION_rts (since
```

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -404,6 +405,41 @@ applyPatchDisableCompileTimeOptimizations ghcFlavor =
...
+          , "        (major1) == 1 && (major2) == 0 && (minor) <= 1)"
+          , "#  else"
+          , "#    define MIN_VERSION_rts(major1,major2,minor) (\\"
+          -- For ghc <= 9.0.1 rts is 1.0.0.
+          , "        (major1) <  1 ||\\"
+          , "        (major1) == 1 && (major2) <  0 ||\\"
+          , "        (major1) == 1 && (major2) == 0 && (minor) <= 0)"
+          , "#  endif"
+          , "#endif" ]
+    -- Write this definition before it's tested on.
```
cocreature, line 434: I don’t understand this. Why do we have to generate this ourselves instead of relying on cabal to produce this macro?

azure-pipelines.yml
shayne-fletcher: we need autoconf so that we can generate prerequisite files
```
  system_ "stack --stack-yaml hadrian/stack.yaml exec -- bash -c ./boot"
  system_ "stack --stack-yaml hadrian/stack.yaml exec -- bash -c \"./configure --enable-tarballs-autodownload\""
  ```
  we were on autocon@2.69. i haven't tried 2.13. i'll give it a go

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: ghc-lib does not express a dependence on rts in it's cabal file

cocreature: Should it? Given this, we clearly do seem to depend on it.

shayne-fletcher: this i don't know. i can try wacking it into the ghc-lib dependencies and removing the macro. let's see.

azure-pipelines.yml
shayne-fletcher: 2.13 exhibits the same sort of errors as 2.71. i think these errors are trying to tell us there is a missing package but what? i don't want to install either of these locally because then i can get myself into a situation where i can't build ghc. fall back on the above fix for now while we work this out.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: depending on rts does the trick

shayne-fletcher:
@cocreature 
- Revised. Don't synthesize the macro instead introduce a dependency of ghc-lib on rts
- autoconf@2.13 exhibits similar errors to 2.71. as i don't have a better fix right now fall back on the one in this PR.

azure-pipelines.yml
shayne-fletcher: heh. i just installed autoconf-2.71 locally (not via brew) and found that i get a lot of warnings from it but that the .boot step succeeds. so, i’m backing up now trying to understand why i felt compelled to change the azure config. i feel like i should be able to revert the azure changes and accept brew’s install of 2.71 or at least not have to install a custom autoconf.

shayne-fletcher: got this worked out. newer versions of the automake toolchain are incompatible with ghc-lib flavors <= 9.0.1. workaround by patching alocal.m4: replace `_AC_PROG_CC_C99` with `AC_PROG_CC_C99`. revert changes to `azure-pipelines.yml`.

shayne-fletcher:
This PR now:
  - Adds `rts` to the `build-depends` in `ghc-lib.cabal`
    - Abandons the earlier patch to `InfoTables.hsc`;
  -  Patches `alocal.m4`:
    -  Replaces `_AC_PROG_CC_C99 ` with `AC_PROG_CC_C99 `
    - Abandons the earlier patch to `azure-pipelines.yml`;
 
 Additionally:
   - Changes to a bash based brew installation method:
    - The older Ruby based version is deprecated;
  - Removes `brew update gmp` from `azure-pipelines.yml`:
    - The action is now unnecessary (has become a no-op). 
  
  Cc @cocreature

cocreature (approved):
Thanks!

shayne-fletcher:
@cocreature if you fancy now tickling the merge button on this i can proceed to rebasing https://github.com/digital-asset/ghc-lib/pull/290 and we'll be outta here!

PR #292, opened 2021-04-11
Sync head; handle platform constants changes; bugfix mini-hlint

shayne-fletcher:
- Fix bug in `mini-hlint` recently introduced: failed to update `Located`  to `LocatedA`;
- Sync to `6974c9e478120f6c4eeb53ebfa935c30cafcdf8e` @ https://gitlab.haskell.org/ghc/ghc.git;
- Changes to adapt to a change in the handling of platform constants  [in this MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5427):
  - The key change is that `DerivedConstants.h`  is now read at runtime (in the case of `mini-compile`) (before now we read the settings file `platformConstants` - this is no longer generated and not to be expected in the target ghc-lib package);
  - When looking for the header so as to read it, we require the `ghc-lib` generated version be the one found;
  - The compiler include path is tested (successfully) before the ghc-lib specific include paths so the above requirement is violated;
  - There's precedent for this problem with a standard soution: rename the generated file `GhclibDerivedConstants.h` and change the ghc-lib sources to look for that instead of `DerivedConstants.h`;
- Raised upstream issue: https://gitlab.haskell.org/ghc/ghc/-/issues/19687:
  - Added `applyPatchGhcPrelude` as a workaround until the issue is resolved;
  - Turns out my issue was a duplicate of https://gitlab.haskell.org/ghc/ghc/-/issues/19683 so closed it.


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -1002,7 +1052,7 @@ generatePrerequisites ghcFlavor = do
           , "--build-root=ghc-lib"
         ] ++
         (if ghcFlavor >= Ghc901 then ["--bignum=native"] else ["--integer-simple"]) ++
-        ghcLibParserExtraFiles ghcFlavor ++ map (dataDir </>) dataFiles
+        ghcLibParserExtraFiles ghcFlavor [] ++ map (dataDir </>) (dataFiles ghcFlavor)
```
cocreature, line 1055: This seems a bit backwards. Instead of defining source and target, you only define a function that applies a substittion to the source and then apply it to an empty substitution in the places where you need the source. Can we extract the source lists to a top-level definition and then use that here directly? I think that’d be clearer than what we do now.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -854,8 +892,14 @@ generateGhcLibCabal ghcFlavor = do
...
         , "bug-reports: https://github.com/digital-asset/ghc-lib/issues"
         , "data-dir: " ++ dataDir
-        , "data-files:"] ++ indent dataFiles ++
-        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor \\ [hadrianGeneratedRoot ghcFlavor </> "ghcversion.h"]) ++
+        , "data-files:"] ++ indent (dataFiles ghcFlavor) ++
+        [ "extra-source-files:"] ++
+          indent (ghcLibExtraFiles ghcFlavor
+                    [ (hadrianGeneratedRoot ghcFlavor </> "ghcversion.h", Nothing)
+                    , (hadrianGeneratedRoot ghcFlavor </> "DerivedConstants.h", Just $ hadrianGeneratedRoot ghcFlavor </> "GhclibDerivedConstants.h")
+                    , ("compiler" </> "HsVersions.h", Just $ "compiler" </> "GhclibHsVersions.h")
```
cocreature, line 900: Seems identical to the list below, let’s factor it out.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -854,8 +892,14 @@ generateGhcLibCabal ghcFlavor = do
         , "homepage: https://github.com/digital-asset/ghc-lib"
         , "bug-reports: https://github.com/digital-asset/ghc-lib/issues"
         , "data-dir: " ++ dataDir
-        , "data-files:"] ++ indent dataFiles ++
-        [ "extra-source-files:"] ++ indent (ghcLibExtraFiles ghcFlavor \\ [hadrianGeneratedRoot ghcFlavor </> "ghcversion.h"]) ++
+        , "data-files:"] ++ indent (dataFiles ghcFlavor) ++
+        [ "extra-source-files:"] ++
+          indent (ghcLibExtraFiles ghcFlavor
+                    [ (hadrianGeneratedRoot ghcFlavor </> "ghcversion.h", Nothing)
+                    , (hadrianGeneratedRoot ghcFlavor </> "DerivedConstants.h", Just $ hadrianGeneratedRoot ghcFlavor </> "GhclibDerivedConstants.h")
```
cocreature, line 899: Was this just broken before? Afaict, we used `DerivedConstants.h` instead of `GhclibDerivedConstants.h`.

shayne-fletcher: we did use just `DerivedConstants.h` before. this means either it didn't matter if the build compiler version got substituted for it when building `ghc-lib` or, it didn't before get referenced during the building of `ghc-lib` as it does now. i think the latter is more probable.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: ok

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: ok. to address these two points,  `ghcLibParserExtraFiles` and `ghcLibExtraFiles` will no longer have substituion logic. apply substitutions in new function `performExtraFilesSubsitutions` and only when there are substitutions to apply.

shayne-fletcher:
@cocreature revised & pushed. thanks!

cocreature (approved):

PR #295, opened 2021-05-01
Sync head

shayne-fletcher:
- Sync to `e61d2d47c4942c829ec98016be5aa5ae36982524`;
- Update `calcParserModules` as a result of https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5627.

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -328,7 +328,7 @@ calcParserModules ghcFlavor = do
   -- The idea here is harvest from lines like
   -- 'compiler/prelude/PrelRules.o : compiler/prelude/PrelRules.hs',
   -- just the module name e.g. in this example, 'PrelRules'.
-      -- Stip comment lines.
+      -- Strip comment lines.
   let depends = filter (not . isPrefixOf "#") (lines buf)
       -- Restrict to Haskell source file lines.
       moduleLines = filter (isSuffixOf ".hs") depends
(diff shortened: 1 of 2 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -342,12 +342,19 @@ calcParserModules ghcFlavor = do
...
-      -- The modules in this list would by default end up in
-      -- ghc-lib. We intervene so that rather, they go into
-      -- ghc-lib-parser.
+      -- The modules in this list elude being listed in
+      -- '.parser-depends' but are required by ghc-lib-parser. We
+      -- intervene and patch things up.
       extraModules =
-        [ "GHC.Driver.Config" | ghcFlavor > Ghc901 ] ++
-        [ "GHC.Parser.Errors.Ppr" | ghcFlavor > Ghc901 ] ++
+        [ x | ghcFlavor > Ghc901
```
cocreature, line 349: why do we need this for 9.2 as well? I would have expected only HEAD to change

shayne-fletcher: you are quite right. thanks. stand by, fix on its way

shayne-fletcher: should be sorted now @cocreature

shayne-fletcher:
any chance of an `/azp run` please?

cocreature (approved):
Thanks!

PR #299, opened 2021-05-27
Sync head

shayne-fletcher:
- Sync `a3665a7aa5db8a77809b8e2246b8cd7eee86935c`
- Update example error handling adapting to [Support new parser types in GHC](https://gitlab.haskell.org/ghc/ghc/-/commit/cdbce8fc22448837e53515946f16e9571e06f412)

### DIFF
--- a/examples/strip-locs/src/Main.hs
+++ b/examples/strip-locs/src/Main.hs
@@ -15,10 +15,15 @@ module Main (main) where
 #  define GHC_921
 #elif MIN_VERSION_ghc_lib_parser(9,0,0)
 #  define GHC_901
-#elif MIN_VERSION_ghc_lib_parser(8,10,1)
+#elif MIN_VERSION_ghc_lib_parser(8,10,0)
 #  define GHC_8101
 #endif
 
+#if defined (GHC_MASTER)
+import "ghc-lib-parser" GHC.Driver.Errors.Types
+import "ghc-lib-parser" GHC.Types.Error hiding (getMessages)
+import qualified "ghc-lib-parser" GHC.Types.Error (getMessages)
+#endif
 #if defined (GHC_MASTER) || defined (GHC_921)
 import "ghc-lib-parser" GHC.Driver.Config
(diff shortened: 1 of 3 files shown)


examples/strip-locs/src/Main.hs
```diff
@@ -15,10 +15,15 @@ module Main (main) where
...
 #  define GHC_901
-#elif MIN_VERSION_ghc_lib_parser(8,10,1)
+#elif MIN_VERSION_ghc_lib_parser(8,10,0)
 #  define GHC_8101
 #endif
 
+#if defined (GHC_MASTER)
+import "ghc-lib-parser" GHC.Driver.Errors.Types
+import "ghc-lib-parser" GHC.Types.Error hiding (getMessages)
+import qualified "ghc-lib-parser" GHC.Types.Error (getMessages)
```
cocreature, line 25: why do we need to import it qualified?

cocreature (approved):
lgtm except for the slightly odd qualified import

examples/strip-locs/src/Main.hs
shayne-fletcher: two modules now give definitions to the common identifier`getMessages`, `GHC.Parser.Lexer` and `Ghc.Types.Error`. i am preferring to leave the `GHC.Parser.Lexer` usages unqualified and insist on qualification for the `GHC.Types.Error.getMessages` when that's the function meant. hence the `import ... hiding(getMessages)` then `import qualified ... (getMessages)` maneuver. do you have a suggestion for a better phrasing?

shayne-fletcher: fyi: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5580#note_355102

cocreature: I’d probably remove the `hiding` part so unqualified usages will give an ambiguous error but this sems fine as well. Thanks for the explanation!

PR #300, opened 2021-06-01
Add a mini-hlint expect test

shayne-fletcher:
- Sync to `6db8a0f76ec45d47060e28bb303e9eef60bdb16b`
- Get a windows `ghc-master` ghc-8.10.4 build going  in CI 
- Add a `--no-checkout` argument to `CI.hs`
  - When it's passed:
    - Don't `rm -rf ghc` on startup
    - Don't `git clone` just refresh existing clone
- Various`mini-hlint` improvements
  - Fix various broken things
  - Write a golden test-suite for it
    - Run that rather than mini-hlint directly in CI

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -27,12 +27,13 @@ main = do
             <> Opts.progDesc "Build ghc-lib and ghc-lib-parser tarballs."
             <> Opts.header "CI - CI script for ghc-lib"
           )
-    Options { ghcFlavor, stackOptions } <- Opts.execParser opts
-    version <- buildDists ghcFlavor stackOptions
+    Options { ghcFlavor, noGhcCheckout, stackOptions } <- Opts.execParser opts
+    version <- buildDists ghcFlavor noGhcCheckout stackOptions
     putStrLn version
 
 data Options = Options
     { ghcFlavor :: GhcFlavor
+    , noGhcCheckout :: Bool
     , stackOptions :: StackOptions
     } deriving (Show)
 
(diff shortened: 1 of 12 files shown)


CI.hs
```diff
@@ -217,26 +224,53 @@ buildDists
...
+            Ghc882 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.2-release"
+            Ghc883 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.3-release"
+            Ghc884 -> cmd "cd ghc && git fetch --tags && git checkout ghc-8.8.4-release"
+            Da { mergeBaseSha, patches, upstream } -> do
+                cmd $ "cd ghc && git fetch --tags && git checkout " <> mergeBaseSha
+                -- Apply Digital Asset extensions.
+                cmd $ "cd ghc && git remote add upstream " <> upstream
+                cmd "cd ghc && git fetch upstream"
+                cmd $ "cd ghc && git -c user.name=\"Cookie Monster\" -c user.email=cookie.monster@seasame-street.com merge --no-edit " <> unwords patches
+            GhcMaster hash -> cmd $ "cd ghc && git checkout " ++ hash
```
cocreature, line 272: This is duplicating 95% of the logic but the only difference is that in one case we call `git clone` (and for some reason `git fetch --tags`?) Can we share the 95% by calling the `git clone` step separately beforehand?

cocreature (approved):
Thanks!

CI.hs
shayne-fletcher: absolutely. done.

shayne-fletcher: mind giving this an `/azp run` please?

shayne-fletcher: oops. introduced an error. sorry. fixed now.

PR #329, opened 2021-10-30
ghc-9.2.1-release

shayne-fletcher:
- update the `ghc-flavor 9.2.1` tag to refer to the released version

PR #330, opened 2021-10-31
package syntax fix

shayne-fletcher:
- requires a rebase on https://github.com/digital-asset/ghc-lib/pull/329 before landing;
- in mini-compile, when the build-compiler is >= 9.2.1, refer to the parser modules re-exported from `ghc-lib` as in pkg `"ghc-lib"` not `"ghc-lib-parser"`.

   i haven't looked into the history but this would appear to be an upstream bug-fix
  - in https://github.com/digital-asset/ghc-lib/pull/329 we worked around 9.2.1's exhibited behavior by adding `ghc-lib-parser` to the `build-depends` section in `mini-compile.cabal`. this fix allows to restore its absence.

### DIFF
--- a/examples/mini-compile/mini-compile.cabal
+++ b/examples/mini-compile/mini-compile.cabal
@@ -16,8 +16,6 @@ executable mini-compile
                      , directory
                      , extra
                      , ghc-lib
--- TODO: investigate. shouldn't be necessary but 9.2.1 fails without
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
@cocreature this has now been rebased with thanks. might we give this an `/azp run` please?

cocreature (approved):
Thanks!

PR #331, opened 2021-11-10
new header ghc-llvm-version.h

shayne-fletcher:
in this sync with head we find a new header `ghc-llvm-version.h`


cocreature (approved):
Thanks!

PR #333, opened 2021-11-14
Support Aeson 2

josephcsible:


josephcsible:
For now, this is just useful for building outside of stack. Once we update to a stack resolver that has aeson >=2, then it should be merged.

cocreature:
> For now, this is just useful for building outside of stack. Once we update to a stack resolver that has aeson >=2, then it should be merged.

how hard is it to support both at the same time so we don’t have to block merging on this? I expect it’ll take a while for the ecosystem to catch up to aeson 2.0 everywhere.

cocreature:
I believe #334 covers the code changes in a backwards compatible way so if we change this to only adjust the bound we can merge it.

shayne-fletcher:
still interested in doing something on this @josephcsible? would be delighted if you are else if you need i can pick it up, let me know!

shayne-fletcher:
@garyverhaegen-da let's close this please

PR #335, opened 2021-11-14
integrate ghc-9.2.1 builds

shayne-fletcher:
this PR gets ghc-9.2.1 builds in ci

- update `stack-exact.yaml` default resolver to `ghc-9.2.1` and update packages/package versions
- the 9.2.1 builds do not execute the `mini-hlint` golden tests
  - this is a stack issue; the next upgrade will unblock re-enabling them
- `ghc-lib-gen` for 9.2.1
  - `-fwarn-incomplete-uni-patterns` seems to have become the default; disable it
  - aeson got a super major version upgrade: notes on migration [here](https://github.com/haskell/aeson/issues/881)
- azure configuration updates:
  - add 9.2.1 builds and remove some older ones
  - move 9.0.1 builds off `stack-exact.yaml` and onto `stack-901.yaml` (works around a happy problem encountered when used with `stack.yaml`)
  - swap out some 8.10.4 builds for 8.10.7 ones

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,81 +41,79 @@ strategy:
     # +---------+-----------------+------------+
     # | OS      | ghc-lib flavour | GHC        |
     # +=========+=================+============+
+    # | linux   | ghc-master      | ghc-9.2.1  |
+    # | macOS   | ghc-master      | ghc-9.2.1  |
+    # +---------+-----------------+------------+
+    linux-ghc-master-9.2.1:
+      image: "ubuntu-latest"
+      mode: "--ghc-flavor ghc-master"
+      resolver: "ghc-9.2.1"
+      stack-yaml: "stack-exact.yaml"
+    mac-ghc-master-9.2.1:
+      image: "macOS-latest"
+      mode: "--ghc-flavor ghc-master"
+      resolver: "ghc-9.2.1"
+      stack-yaml: "stack-exact.yaml"
+
+    # just can't get a build-plan at this time. issues with extra,
+    # directory, process & time
+
+    # windows-ghc-master-9.2.1:
+    #   image: "windows-latest"
+    #   mode: "--ghc-flavor ghc-master"
+    #   resolver: "ghc-9.2.1"
+    #   stack-yaml: "stack-exact.yaml"
+
+    # +---------+-----------------+------------+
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -38,84 +38,82 @@ strategy:
     #      | HEAD         |  >= 8.10.1  | (since 2021/03/10, https://gitlab.haskell.org/ghc/ghc/-/commit/0a709dd9876e40c19c934692415c437ac434318c)
     #      +--------------+-------------+
 
+    # +---------+-----------------+------------+
+    # | OS      | ghc-lib flavour | GHC        |
+    # +=========+=================+============+
+    # | linux   | ghc-master      | ghc-9.2.1  |
+    # | macOS   | ghc-master      | ghc-9.2.1  |
+    # +---------+-----------------+------------+
+    linux-ghc-master-9.2.1:
```
cocreature, line 47: I’m a bit worried that we just keep extending the number of CI jobs. I think we should start thinking about which ones we can drop reasonably. Maybe 8.8 can go (excluding the daml fork), maybe we can limit cross-platform tests to only the most recent one.

Nothing for this PR but I think we have to address this soonish.

examples/mini-hlint/test/Main.hs
```diff
@@ -41,24 +41,44 @@ runTest flavor f =
...
-                        takeDirectory hsFile </> "MiniHlintTest_fatal_error-ghc-master.hs"
-                      "MiniHlintTest_non_fatal_error.hs" ->
-                        takeDirectory hsFile </> "MiniHlintTest_non_fatal_error-ghc-master.hs"
-                      _ -> hsFile
-                  _ -> hsFile
-            in replaceExtension f $ (if isWindows then ".windows" else "") ++ ".expect"
-    , let genStringAction = stack stackYaml resolver $ "--no-terminal exec -- mini-hlint " ++ hsFile
-  ]
+goldenTests stackYaml@(StackYaml yaml) stackResolver@(Resolver resolver) (GhcFlavor ghcFlavor) hsFiles =
+  -- Note: You'll get very confused if you load expect files into your
```
cocreature, line 45: why not change the tests to write out a trailing newline?

cocreature (approved):
Thanks!

PR #339, opened 2021-11-25
remove applyPatchPrelude

shayne-fletcher:
the fix for bits bounds was backported to 9.2.1 in `1571772e37b587495dea7f2e0c575c70141027d8` making `applyPatchPrelude` a no-op. remove.


cocreature (approved):
Thanks!

PR #342, opened 2021-12-21
patch parser

shayne-fletcher:
- [Hlint issue](https://github.com/ndmitchell/hlint/issues/1333)
- [GHC issue](https://gitlab.haskell.org/ghc/ghc/-/issues/20846)
- [GHC MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7226)

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -11,7 +11,8 @@
 #endif
 
 module Ghclibgen (
-    applyPatchHeapClosures
+    applyPatchParser
+  , applyPatchHeapClosures
   , applyPatchAclocal
   , applyPatchHsVersions
   , applyPatchGhcPrim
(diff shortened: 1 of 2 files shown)


cocreature (commented):
I’m not that keen on patching this here. I’d like to have ghc-lib stay as close to the upstream source code as possible and only patch for ghc-lib specific stuff that isn’t going to get fixed upstream.

ndmitchell:
Agree with @cocreature that staying as close to upstream as possible is a good thing. I think we should try and get a patch into GHC 9.2.2, and if the release of 9.2.2 isn't imminent (e.g. they don't consider a bug in the API worth a point release for) then patching that into ghc-lib would make a lot of sense. However, if there is going to be a long lag with GHC HQ, I think doing something temporary is more reasonable.

cocreature:
while it took longer than expected 9.2.2 is out now so closing this.

Issue #354, opened 2022-03-16
ghc-lib-parser-9.2.1.* fails to compile with GHC-9.2.2


phadej:
```haskell
FFIClosures_ProfilingDisabled.hsc: In function ‘main’:
FFIClosures_ProfilingDisabled.hsc:81:16: error: ‘BlockedOnIOCompletion’ undeclared (first use in this function); did you mean ‘BlockedOnDoProc’?
```

all three versions

- ghc-lib-parser-9.2.1.20220109
- ghc-lib-parser-9.2.1.20211101
- ghc-lib-parser-9.2.1.20211030

seems to be affected. Seem to compile fine with `ghc-9.2.1` though. Please make revisions.

phadej:
That is somewhat irrelevant as these releases are broken with GHC-9.2.2.

I'd like a situation where a constraint solver tells me that these versions are not usable with GHC-9.2.2, instead of me trying to figure out which release works. E.g. if I have a freeze file which worked with GHC-9.2.1.

shayne-fletcher:
we generally proxy with base bounds to prevent a build plan but that's not an option in this case

phadej:
Why it isn't? GHC-9.2.1 and GHC-9.2.2 come with different `base`s (`base-4.16.0.0` and `base-4.16.1.0`)

shayne-fletcher:
have patched the bounds for these 3 on hackage so that now we get a failed build plan as desired e.g.
```
[__1] rejecting: base-4.16.1.0/installed-4.16.1.0 (conflict: ghc-lib-parser =>
base>=4.14 && <4.16.1)
```
will put up a patch up for ghc-lib-gen for any further 9.2.1 releases

phadej:
thanks!

PR #355, opened 2022-03-16
ghc-flavor=ghc-9.2.1 requires base < 4.16.1

shayne-fletcher:
Tighten the upper bound on base for `ghc-flavor=ghc-9.2.1` to `4.16.1` which excludes a build plan where `ghc-9.2.2` is the build compiler. Fixes https://github.com/digital-asset/ghc-lib/issues/354.


sofiafaro-da (approved):
Thanks!

PR #356, opened 2022-03-20
Don't pass -haddock ghc option to stack builds

shayne-fletcher:
It causes breakages with `HEAD` (like [this one just committed](https://gitlab.haskell.org/ghc/ghc/-/commit/97db789eec7a49c3ec30a83666720221c26d8f9e)) too regularly and with [this landing](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7762) it will get checked upstream now.


sofiafaro-da (approved):
Thanks!

shayne-fletcher:
ping

PR #370, opened 2022-05-20
no longer require fake llvm config

shayne-fletcher:
- `LllvmConfig` is no longer stored in `DynFlags` (reference [this commit](https://gitlab.haskell.org/ghc/ghc/-/commit/ef3c8d9e077a1d4ede0724075489fb1f12afa3f9)). fix the examples accordingly
- the 9.4.1 (alpha) branch is looking broken right now so disable that flavor of build for now
- da-ghc builds also removed as they are broken presumbably by https://github.com/digital-asset/ghc/pull/119

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -69,7 +69,7 @@ data GhcFlavor = Ghc941
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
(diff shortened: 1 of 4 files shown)


shayne-fletcher:
cc @moisesackerman-da @sofiafaro-da @garyverhaegen-da

PR #372, opened 2022-05-25
upgrade to ghc-9.0.2

shayne-fletcher:
ghc-9.4.1 is coming (late july). it's time to clean up some older builds and move the default resolver to point at a ghc-9.0.2 build (minimum bootstrap for 9.4.1 and master is the 9.0 series).


shayne-fletcher:
cc @garyverhaegen-da @moisesackerman-da @sofiafaro-da

azure-pipelines.yml
```diff
@@ -73,65 +75,21 @@ strategy:
...
       image: "macOS-latest"
       mode: "--ghc-flavor ghc-master"
-      resolver: "nightly-2021-11-14"
-      stack-yaml: "stack-901.yaml"
-    mac-ghc-master-8.10.7:
-      image: "macOS-latest"
-      mode: "--ghc-flavor ghc-master"
-      resolver: "lts-18.16"
-      stack-yaml: "stack.yaml"
-    windows-ghc-master-8.10.7:
```
garyverhaegen-da, line 114: macOS and Linux have been cleaned up a bit, but Windows seems to have just disappeared. Was that intentional? If so, would you mind sharing a bit more context on why Windows is no longer desired here?

shayne-fletcher: sure. it was intentional. the last remaining windows build was an 8.10.7 and moving to 9.0.2 broke it. for various reasons it's been impossible (for reasons beyond our control) to obtain a windows build since then. i have zero appetite to sink any more hours into it. i am disinclined therefore to continue testing windows builds.

shayne-fletcher: note however that ghc-lib builds themselves (as opposed to the production of them) continue to be tested incidentally by ghc-lib-parser-ex and hlint.

shayne-fletcher: @garyverhaegen-da

garyverhaegen-da: It's been a while since I looked at this project; can you double-check if this story I remember is vaguely true?

This project is a Haskell program that packages GHC source code in a way that can be used as a library. The result of running this program is a tarball of Haskell source code that represents a subset of the GHC source tree. Which platform that tarball has been produced on is pretty much irrelevant, since they're just source files.

We keep testing on multiple platforms only to allow for development _on this project_ on multiple platforms, and therefore if there's no one working on this project that uses Windows, running CI on Windows is useless.

The source tarballs can be used on any platform that has a Haskell compiler, so dropping Windows here doesn't prevent e.g. the daml builds from building the daml compiler, using the tarballs produced by running ghc-lib on Linux, from building on Windows.

Is this in the right ballpark?

shayne-fletcher: > This project is a Haskell program that packages GHC source code in a way that can be used as a library. The result of running this program is a tarball of Haskell source code that represents a subset of the GHC source tree. Which platform that tarball has been produced on is pretty much irrelevant, since they're just source files.

correct.

> We keep testing on multiple platforms only to allow for development _on this project_ on multiple platforms, and therefore if there's no one working on this project that uses Windows, running CI on Windows is useless.

correct. 

> The source tarballs can be used on any platform that has a Haskell compiler, so dropping Windows here doesn't prevent e.g. the daml builds from building the daml compiler, using the tarballs produced by running ghc-lib on Linux, from building on Windows.

correct.

> 
> Is this in the right ballpark?

yes.

shayne-fletcher:
> Thanks!

no, thank-you gary on two counts (1) your timely responses & actions and (2) your willingness to understand. much gratitude my friend.

azure-pipelines.yml
```diff
@@ -168,33 +126,6 @@ strategy:
...
-    # windows-da-ghc-8.8.1-8.8.1:
-    #   image: "windows-latest"
-    #   mode: "--da"
-    #   resolver: "nightly-2020-01-08"
-    #   stack-yaml: "stack.yaml"
-    # mac-da-ghc-8.8.1-8.8.1:
-    #   image: "macOS-latest"
-    #   mode: "--da"
-    #   resolver: "nightly-2020-01-08"
-    #   stack-yaml: "stack.yaml"
```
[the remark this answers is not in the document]
garyverhaegen-da: Yes, but not the Windows one I'd say.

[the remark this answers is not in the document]
shayne-fletcher: @cbley-da it's up to digital-asset of course but i have no objection to it!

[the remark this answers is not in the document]
shayne-fletcher: > Yes, but not the Windows one I'd say.

the one on windows for da-ghc-8.8.1 should be fine!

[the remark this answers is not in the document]
shayne-fletcher: (however i would be just as happy if we didn't re-enable that one; "in for a penny in for a pound" as it were :) )

PR #374, opened 2022-05-26
tweak mini-compile cabal file

shayne-fletcher:
i glossed over this in https://github.com/digital-asset/ghc-lib/pull/372 but 9.0.2 seems to be broken in the following way: `ghc-lib` re-exports `ghc-lib-parser` so it shouldn't be necessary to `build-depend` on `ghc-lib-parser` when you `build-depend` on on `ghc-lib`.  i worked around 9.0.2 appearing to violate this rule by adding an unconditional `build-depend` of mini-compile on `ghc-lib-parser`. i'm circling back in this PR and making it conditional on `>=9.0.2 && <9.3` (i'll also mention that i manually ran this patch with ghc-9.0.1 and it was ok  there too as i expected)

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -69,7 +69,7 @@ data GhcFlavor = Ghc941
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current = "ffbe28e56aa382164525300fbc32d78eefd95e7d" -- 2022-05-23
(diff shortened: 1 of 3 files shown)


shayne-fletcher:
@garyverhaegen-da a.u.b.!

garyverhaegen-da:
Why is this now closed? Is it included in another PR?

shayne-fletcher:
@garyverhaegen-da closed because it turned out not to work according to my expectations so abandoned for the moment

PR #377, opened 2022-06-02
Allow specifying extra CPP flags to include in generated .cabal files

akrmn:
ghc-lib PR: (here)
da-ghc PR: https://github.com/digital-asset/ghc/pull/123
daml PR: https://github.com/digital-asset/daml/pull/14074

PR #380, opened 2022-06-03
Fix DA CI

akrmn:
As @shayne-fletcher noticed (https://github.com/digital-asset/ghc-lib/pull/377#issuecomment-1145744271), the change from https://github.com/digital-asset/ghc/pull/123 & https://github.com/digital-asset/ghc-lib/pull/377 broke CI for the DA build. It didn't break on https://github.com/digital-asset/ghc-lib/pull/377 because https://github.com/digital-asset/ghc/pull/123 was merged after it. The fix is to drop the now-useless patch https://github.com/digital-asset/ghc/tree/da-unit-ids-8.8.1 and instead call `ghc-lib-gen` with the new `--cpp=-DDAML_PRIM` option for the DA flavor.

I also took the opportunity to extract the `Da` record variant of `GhcFlavor` into its own type, as well as replace all uses of "flavour" with the more frequent (in the repo) "flavor".

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -64,8 +64,16 @@ data GhcFlavor = Ghc941
                | Ghc882
                | Ghc883
                | Ghc884
-               | Da { mergeBaseSha :: String, patches :: [String], flavor :: String, upstream :: String }
+               | Da DaFlavor
                | GhcMaster String
+  deriving (Eq, Show)
+
+data DaFlavor = DaFlavor
+  { mergeBaseSha :: String
+  , patches :: [String]
+  , flavor :: String
+  , upstream :: String
+  }
   deriving (Eq, Show)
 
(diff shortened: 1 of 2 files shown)


akrmn:
> hence 'flavor' is correct here!

😅 ah glad we all converged on it

cocreature (approved):
thanks!

shayne-fletcher:
> > hence 'flavor' is correct here!
> 
> 😅 ah glad we all converged on it

ha, ha, most of the wrong spellings were my fault. trouble is, like the british, we use english spellings in my country (new zealand) too 🤣

akrmn:
> ha, ha, most of the wrong spellings were my fault. trouble is, like the british, we use english spellings in my country (new zealand) too 🤣

haha nah I've probably introduced some too, between living a couple of years in London and being surrounded by the in-between "euro" English

PR #381, opened 2022-06-04
update stack-exact.yaml

shayne-fletcher:
maintenance PR: update package versions in `stack-exact.yaml`.

when a PR for https://github.com/commercialhaskell/stackage-content/issues/103 lands, we'll bump to 9.2.3

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,6 +38,10 @@ strategy:
     # (the general rule is only the last two compiler versions are
     # supported).
 
+    # note(sf, 2022-0604): We should be able to bump this to ghc-9.2.3
+    # soon (see
+    # https://github.com/commercialhaskell/stackage-content/issues/103).
+
     # +---------+-----------------+------------+
     # | OS      | ghc-lib flavor  | GHC        |
     # +=========+=================+============+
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
@garyverhaegen-da

azure-pipelines.yml
```diff
@@ -72,23 +93,6 @@ strategy:
...
-    linux-ghc-9.4.1-9.2.2:
-      image: "ubuntu-latest"
-      mode: "--ghc-flavor ghc-9.4.1"
-      resolver: "ghc-9.2.2"
-      stack-yaml: "stack-exact.yaml"
-    mac-ghc-9.4.1-9.2.2:
-      image: "macOS-latest"
-      mode: "--ghc-flavor ghc-9.4.1"
-      resolver: "ghc-9.2.2"
-      stack-yaml: "stack-exact.yaml"
```
akrmn, line 90: Hi @shayne-fletcher, I don't understand why you changed `linux-ghc-9.4.1-9.2.2` and `mac-ghc-9.4.1-9.2.2` to 
```
      resolver: "nightly-2022-05-27"
      stack-yaml: "stack.yaml"
```
could you explain?

shayne-fletcher: it's a 9.2.2 resolver that's it's changed to so it's equivalent. i'm shaping up to get ghc-9.2.3/stack-exact.yanl going but we're waiting (works locally) but we need the ghc-9.2.3 resolver first

shayne-fletcher: note the key difference: stack. yaml vs stack-exact.yaml. the former has no "extra-deps" it's a curated package collection. the latter is using a resolver that is just a ghc bindist, no packages. all packages are hand specified in the extra-deps.

shayne-fletcher: @moisesackerman-da

shayne-fletcher: here's a little more background to help: the purpose of `stack-exact.yaml` is to get signal on a compiler for which there is a bindist resolver but not yet a curated package resolver. put another way, when there is a new compiler release, and a bindist resolver has been setup for it, we can get signal (it generally takes another month or two after that before there is a curated package resolver). when a curated package resolver becomes available then, happy days we can switch to it. that's the case here with 9.2.2,  a resolver has become available so let's use it. at the same time 9.2.3 is now out, there is neither a bindist or curated package resolver  for that yet available so we're in a holding pattern with that one (as mentioned though, i'm getting early signals locally - see https://github.com/commercialhaskell/stack/issues/5761#issuecomment-1146369807 for info on how to go about doing that). anyway, that's what these changes to the azure pipeline are all about - me going as far as i can for the moment and getting setup for https://github.com/commercialhaskell/stackage-content/issues/103 where we'll use `stack-exact.yaml` with 9.2.3. normal evolution, not a big deal

shayne-fletcher: oh, you might ask why `linux-ghc-master-9.2.2` and `mac-ghc-master-9.2.2` aren't using `nightly-2022-05-27` then. the answer is that i want to see `stack-exact.yaml` get at least some use before 9.2.3 comes along. as soon as the referenced ticket is satisfied, we'll modify `resolver: "ghc-9.2.2"` to `resolver: "ghc-9.2.3"` as indicated by the comment 
```
    # note(sf, 2022-0604): We should be able to bump this to ghc-9.2.3
    # soon (see
    # https://github.com/commercialhaskell/stackage-content/issues/103).
```
sorry for the long winded explanation @moisesackerman-da . hopefully we're all clear now?

garyverhaegen-da: Seems clear to me, thanks for the explanation.

garyverhaegen-da (approved):

azure-pipelines.yml
akrmn: All clear, thank you for the detailed explanation @shayne-fletcher. Also thanks @garyverhaegen-da for approving, I forgot about this PR while looking at other stuff.

PR #384, opened 2022-06-09
add skip-init flag

shayne-fletcher:
this PR delivers a 16% (rel) wall-time reduction in overall execution of CI.hs by making sure to perform initialization and parser-depends steps in CI only once
- add a flag `skip-init` to `ghc-lib-gen`
- in CI.hs `stack exec -- ghc-lib-gen ghc-lib --skip-init`

```
before
      375.78 real      1097.31 user       380.80 sys
after
      323.39 real      1073.91 user       353.09 sys
```

### DIFF
--- a/ghc-lib-gen/src/Main.hs
+++ b/ghc-lib-gen/src/Main.hs
@@ -4,10 +4,12 @@
 
 module Main(main) where
 
-import System.Directory
 import Ghclibgen
 import GhclibgenOpts
+
+import System.Directory
 import Options.Applicative
+import Control.Monad
 
 main :: IO ()
 main = ghclibgen =<< execParser opts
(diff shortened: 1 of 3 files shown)


shayne-fletcher:
@garyverhaegen-da @moisesackerman-da this saves us another minute or so in CI

akrmn (approved):
very nice, thank you!

ghc-lib-gen/src/Main.hs
```diff
@@ -51,6 +56,6 @@ ghclibgen (GhclibgenOpts root target ghcFlavor cppOpts) =
         -- 'hadrian/stack.yaml'.
         generatePrerequisites ghcFlavor
         -- Renamings come after 'generatePrerequisites':
+        applyPatchDerivedConstants ghcFlavor -- Needs DerivedConstants.h
```
akrmn, line 59: out of curiosity, did it make a difference to do this step earlier?

shayne-fletcher: no, not significant i think, i probably just moved it while i was working stuff out. key point is it must follow `generatePrequisites` is all

shayne-fletcher: also, if you don't do a `git update` after renaming it (which is what happens in `applyPatchDerivedConstants` iirc), that step will fail for the lack of it

shayne-fletcher: or something 😂

PR #387, opened 2022-06-13
make sdists of examples

shayne-fletcher:
add a little code to `CI.hs` that produces sdist of `test-utils`, `mini-hlint`, `mini-compile`. although we don't publish these on hackage doing this enables me to treat them uniformly with `ghc-lib-parser`, `ghc-lib` and `hlint` in my local end-to-end tests (takes a step closer to getting them in a state suitable for the hlint repo)


shayne-fletcher:
@garyverhaegen-da @akrmn

PR #394, opened 2022-07-06
workaround vendor filepath hack

shayne-fletcher:
- MR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8516:
  - Temporary hack to vendor in filepath breaks cabal parsing in `ghc-lib-gen`
  - Work around

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -25,6 +25,7 @@ module Ghclibgen (
   , applyPatchNoMonoLocalBinds
   , applyPatchCmmParseNoImplicitPrelude
   , applyPatchHadrianStackYaml
+  , applyPatchTemplateHaskellCabal
   , generatePrerequisites
   , mangleCSymbols
   , generateGhcLibCabal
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
cc @garyverhaegen-da @akrmn @sofiafaro-da @cocreature

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -355,7 +356,7 @@ calcParserModules ghcFlavor = do
           [ "GHC" </> "Parser.hs" | ghcFlavor >= Ghc901] ++
           [ "Parser.hs" | ghcFlavor < Ghc901]
         )
-
```
shayne-fletcher, line 358: surprisingly some redundant whitespace has crept in which is getting removed here incidentally

akrmn (approved):
Work around a workaround :) 
Thanks!

shayne-fletcher:
> Work around a workaround :) Thanks!

@akrmn  😆 

btw, i forget what it's called and i don't have privilege to check but there's a label you can apply so that, now you've approved, if/when the checks pass it will auto-land

moisesackerman-da:
@shayne-fletcher hm I couldn't find the label, but I do have the auto-merge button that does the same, I'll remember to use that next time

PR #396, opened 2022-07-11
add new flag to skip building things

shayne-fletcher:
add `CI.hs` option `--no-builds`. when this switch is provided, `CI.hs` will terminate immediately after it has written ghc-lib package archives `.tar.gz`. that is, the packages themselves and the examples are neither built nor tested. this enables a quick end-to-end testing capability for hlint.


shayne-fletcher:
@akrmn @garyverhaegen-da

garyverhaegen-da (approved):

PR #401, opened 2022-08-08
mini-compile example: unconditionally get parser modules through pkg ghc-lib-parser

shayne-fletcher:
the `mini-compile` preprocessor logic around `GHC_LIB_PARSER_PKG` is malformed, doesn't make any sense and breaks with ghc-9.4.1. just remove it.


garyverhaegen-da (approved):
Thanks!

PR #411, opened 2022-10-02
fix missing cleanup in CI.hs

shayne-fletcher:
add in a missing cleanup action in CI.hs. non-critical

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -69,7 +69,7 @@ data DaFlavor = DaFlavor
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current = "615e22789a04e74d7e02239b4580b95b077c3ae0" -- 2022-09-29
+current = "2209665273135644f1b52470ea2cb53169f2ef91" -- 2022-10-02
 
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):

shayne-fletcher:
@garyverhaegen-da this azure mac thing seems to to be getting worse 😔 is it not affecting your day-to-day builds too?

garyverhaegen-da:
> @garyverhaegen-da this azure mac thing seems to to be getting worse 😔 is it not affecting your day-to-day builds too?

The failure was a rate limit on GitHub API calls, which are per IP address. Maybe the macOS instances provided by Azure are pooled behind fewer IP addresses (using NAT) than other machines?

It's not affecting us because we're not using the Azure-provided macOS machines (or really any Azure-provided machine at this point AFAIK). We do occasionally get rate limit errors too, but only in rare and specific cases.

I've restarted and one of them failed again, but the other got past the GitHub API call point. I'll keep an eye on this and keep retrying for a while.

garyverhaegen-da:
#412 should help with rate limits in the future.

PR #412, opened 2022-10-03
avoid GitHub API

garyverhaegen-da:
The primary motiviation here is to avoid our jobs failing on GitHub API rate limits, but it's also actually cleaner to not depend on Neil's nested script files, I think.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -176,28 +176,13 @@ steps:
   - bash: |
       set -euo pipefail
 
-      case "$(uname)" in
-        Darwin)
-          os=osx;;
-        MINGW64_NT-*|MSYS_NT-*)
-          os=windows;;
-        *)
-          os=linux;;
-      esac
-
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -170,18 +170,53 @@ steps:
...
+        MINGW64_NT-*|MSYS_NT-*)
+          os=windows;;
+        *)
+          os=linux;;
+      esac
+
+      version=$(curl -Is https://github.com/ndmitchell/hlint/releases/latest | grep location: | sed 's|.*/||' | tr -d '\n\r')
+      echo "VERSION: $version"
+
+      ext=$(if [ "windows" = $os ]; then echo 'zip'; else echo 'tar.gz'; fi)
```
cocreature, line 191: why don’t we just skip hlint on everything but linux?

cocreature (approved):
thx!

PR #416, opened 2022-11-20
enable testing with ghc-9.4.3

shayne-fletcher:
this is just a CI update.  ghc-9.4.3 and ghc-9.2.5 resolvers are now listed (see https://github.com/commercialhaskell/stackage-content/blob/master/stack/stack-setup-2.yaml) so enable new builds and clean up old ones

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,92 +41,52 @@ strategy:
 
     # enable when resolver ghc-9.4.3 exists
 
+    # +---------+-----------------+------------+
+    # | OS      | ghc-lib flavor  | GHC        |
+    # +=========+=================+============+
+    # | macOS   | ghc-master      | ghc-9.4.3  |
+    # | windows | ghc-master      | ghc-9.4.3  |
+    # +---------+-----------------+------------+
+    # This pipeline fails at the ghc-9.4.3 install step with
+    # /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
+    # I think this indicates that ubuntu-latest needs an upgrade.
     # linux-ghc-master-9.4.3:
     #   image: "ubuntu-latest"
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):
Thanks!

shayne-fletcher:
the linux ghc-master ghc-9.43 job failed installing the compiler. this did not happen in [pre-testing](https://github.com/shayne-fletcher/ghc-lib/pull/140/checks?check_run_id=9596389173). i believe it's a transient error and if we re-run the job please @garyverhaegen-da it'll pass.

garyverhaegen-da:
Hi @shayne-fletcher,

This doesn't seem to be transient as it keeps happening (4 out of 4 so far). The error message boils down to:
```
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
```
which makes me wonder if perhaps the CI machines are too "old"?

shayne-fletcher:
> Hi @shayne-fletcher,
> 
> This doesn't seem to be transient as it keeps happening (4 out of 4 so far). The error message boils down to:
> 
> ```
> /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
> ```
> 
> which makes me wonder if perhaps the CI machines are too "old"?

oh gosh. yes, see lines 67-72 of https://github.com/digital-asset/ghc-lib/blob/0a56f2578bcdf52e866011e5217bcc917ed6d6a9/azure-pipelines.yml where we've already encountered this problem. i can't tell you why my own build/tests are not exhibiting this problem but oh well, i'll push a patch that disables the configuration for the time being.

garyverhaegen-da:
I can confirm the build nodes currently embed glibc 2.31.

shayne-fletcher:
> I can confirm the build nodes currently embed glibc 2.31.

disabled the relevant pipeline.

PR #418, opened 2022-11-21
bump ubuntu

garyverhaegen-da:


garyverhaegen-da:
Apparently "latest" means 20.04 at the moment. 🤷

cc @shayne-fletcher

shayne-fletcher:
@garyverhaegen-da i always thought `ubuntu-latest` meant the most recent but evidently that's not the case and `ubuntu-22.04` is "more recent" than `ubuntu-latest`? ok, then, works for me!

akrmn (approved):
Thanks Gary!

I think we should be able to switch back to `ubunut-latest` once https://github.com/actions/runner-images/issues/6399 is closed, when it will be identical to `ubuntu-22.04`,

> ### Target date
>
> This change will be rolled out over a period of several weeks beginning in October, 3. We plan to complete the migration by December, 1.

PR #423, opened 2023-01-15
ghc-9.6.1

shayne-fletcher:
ghc-9.6.1 is in alpha. add support & let's start getting signal.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -51,6 +51,7 @@ data StackOptions = StackOptions
 
 data GhcFlavor = Da DaFlavor
                | GhcMaster String
+               | Ghc961
                | Ghc944 | Ghc943 | Ghc942 | Ghc941
                | Ghc925 | Ghc924 | Ghc923 | Ghc922 | Ghc921
                | Ghc902 | Ghc901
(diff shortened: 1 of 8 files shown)


shayne-fletcher:
cc @garyverhaegen-da @akrmn

CI.hs
```diff
@@ -441,6 +446,7 @@ buildDists
     let ghcOpts = case ghcFlavor of Da {} -> ghcOptionsOpt ghcOptions;  _ -> ghcOptionsWithHaddock ghcOptions
     stack $ "--no-terminal --interleaved-output build " ++ ghcOpts ++ " ghc-lib-parser"
     stack $ "--no-terminal --interleaved-output build " ++ ghcOptionsOpt ghcOptions ++ " ghc-lib"
+#if 0
```
akrmn, line 449: is `ghc-lib-test-mini-hlint` broken on 9.6.1? if so, I'd say we should exclude just that version explicitly, rather than disabling it entirely

shayne-fletcher: oops! good catch! sorry, unintended change. i fixed and pushed. unfortunately, this will mean needing to `/azp run` again. sorry!

shayne-fletcher: argh. re-enabling has reintroduced what i tricked myself into believing was a solved problem. i'll have to work something out. putting this back in draft for now.

akrmn: thank you @shayne-fletcher!

shayne-fletcher:
this problem would appear to be issue https://github.com/commercialhaskell/stack/issues/5866.

CI.hs
shayne-fletcher: this problem would appear to be issue https://github.com/commercialhaskell/stack/issues/5866.

shayne-fletcher:
note to self. technically ghc-9.6.1 source should compile with ghc-9.2.1 but it doesn't, bytestring needs to be >= 0.11.3, it ships with 0.11.0 and the first compiler that has 0.11.3 is 9.2.2.

PR #425, opened 2023-01-22
update stack-exact.yaml

shayne-fletcher:
ghc-9.2.2 resolver & the most recent compatible packages

### DIFF
--- a/stack-exact.yaml
+++ b/stack-exact.yaml
@@ -7,7 +7,7 @@
 # You can add --resolver xxx to the above command to have the effect
 # of overriding the choice of compiler if desired.
 
-resolver: ghc-9.2.4
+resolver: ghc-9.2.2
 
 ghc-options:
   # Try to be quick.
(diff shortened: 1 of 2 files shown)


stack-exact.yaml
```diff
@@ -7,7 +7,7 @@
 # You can add --resolver xxx to the above command to have the effect
 # of overriding the choice of compiler if desired.
 
-resolver: ghc-9.2.4
+resolver: ghc-9.2.2
```
akrmn, line 10: it seems stackage [lts-20.8](https://www.stackage.org/lts-20.8) supports ghc 9.2.5, so why not

```suggestion
resolver: ghc-9.2.5
```
?

akrmn (approved):
LGTM other than question about resolver version

stack-exact.yaml
shayne-fletcher: i chose 9.2.2 because its the minimum supported build compiler for ghc-9.6.1 flavored builds.

shayne-fletcher:
looks like we got a bit hung up on windows da-8.81 @akrmn. hopefully it's transient!

garyverhaegen-da:
> looks like we got a bit hung up on windows da-8.81 @akrmn. hopefully it's transient!

Restarted. 🤞

stack-exact.yaml
akrmn: gotcha! thanks for the explanation

PR #430, opened 2023-02-06
simplify mini-hlint tests

shayne-fletcher:
the mini-hlint golden tests were a nuisance to maintain. this replaces them with something equivalent but much simpler.


shayne-fletcher:
looking good @garyverhaegen-da. two more to follow for now.

PR #432, opened 2023-02-07
stack-yaml should use resolver >= ghc-9.4

shayne-fletcher:
give  `stack.yaml` a 9.4 default resolver so that `stack runhaskell --package extra --package optparse-applicative CI.hs -- --ghc-flavor ghc-master` works


akrmn (approved):
thanks @shayne-fletcher!

PR #442, opened 2023-03-10
refactor gitCheckout

shayne-fletcher:
this is just a tiny refactor that makes CI.hs ever so slightly better.


shayne-fletcher:
cc @garyverhaegen-da @moisesackerman-da

garyverhaegen-da (approved):

PR #447, opened 2023-04-22
patch out semaphore-compat

shayne-fletcher:
implementation of the ghc job server protocol in [this commit](https://gitlab.haskell.org/ghc/ghc/-/commit/5c8731244bc13a3d813d2a4d53b3188b28dc8355) breaks a lot of ghc-lib builds.

for the time being, to not break those ghc-flavor ghc-9.8 builds, we need to avoid ghc-lib depending on the newly created library semaphore-compat. this PR implements a patch that does that.

PR #451, opened 2023-05-06
patch reworked plugin initialization

shayne-fletcher:
- this [rework plugin initialization commit](https://gitlab.haskell.org/ghc/ghc/-/commit/18a7d03d46706d2217235d26a72e6f1e82c62192) introduces further changes to 'GHC/Driver/Make.hs' in addition to the changes that gave rise to [this PR](https://github.com/digital-asset/ghc-lib/pull/447) so update the patch accordingly
- additionally, [this commit](https://gitlab.haskell.org/ghc/ghc/-/commit/983ce55815f2dd57f84ee86eee97febf7d80b470) gives rise to a type mismatch between the package template-haskell and the file `Language.Haskell.TH.Syntax` that appears in ghc-lib-parser (i tried various other approaches to work around this and found out some interesting stuff but for now this keeps things going & appears to be the least thing we can do)
- there's one more thing that i do in this PR which is to `rm $HOME/.stack/setup-exe-src/*` in `CI.hs` (and a related `git remote remove upstream` in the `--no-checkout` case). these things should have no bearing on CI but i worked out enable me to test `--da` builds locally (which i haven't been able to do since forever)

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -71,7 +71,7 @@ data DaFlavor = DaFlavor
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current = "4eaf2c2a7682fa9933261f5eb25da9e2333c9608" -- 2023-04-27
+current = "1e6861dd8612d8f71cc4635c4f73f84f316a6c7e" -- 2023-05-10
 
 -- Command line argument generators.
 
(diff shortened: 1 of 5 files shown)


shayne-fletcher:
cc @akrmn @dylant-da @garyverhaegen-da

shayne-fletcher:
cc @moisesackerman-da @dylant-da @garyverhaegen-da ping 😄

CI.hs
```diff
@@ -326,14 +326,21 @@ buildDists
     --      git fetch --tags && git submodule update --init --recursive
     -- and it won't be deleted between runs.
     if noGhcCheckout then
-      cmd "cd ghc && git clean -xdf && git submodule foreach git clean -xdf && git submodule foreach git checkout . && git checkout ."
+      cmd "cd ghc && git remote remove upstream || true && git clean -xdf && git submodule foreach git clean -xdf && git submodule foreach git checkout . && git checkout ."
```
akrmn, line 329: I'm not sure about the relative precedence of `||` vs `&&` in bash, iiuc you're trying to remove `upstream` but don't care if it's not there to begin with, but I think we should just split this into a few lines

akrmn (approved):
LGTM, thank you @shayne-fletcher! Just one comment about bash, of all things :)

CI.hs
shayne-fletcher: @akrmn  i'll follow up with a new diff that does that. mind landing this one as is for now please? (it's the most critical of the set you've reviewed today)

shayne-fletcher: never mind @akrmn! force pushed a revision that splits up and rebases on master. let's /azp run this one more time and assuming it's good, get it landed. thanks!

PR #456, opened 2023-05-24
bound applyPatchGhcPrim

shayne-fletcher:
This patch affects flavors >= 9.0 with build compilers < 9.0.2 from which it follows flavors > 9.4 should just skip past this.


shayne-fletcher:
@akrmn one more please 🙏

akrmn (approved):
LGTM, thanks @shayne-fletcher!

PR #459, opened 2023-06-07
improve the discovery method of ghc-lib-parser source files

shayne-fletcher:
after the commits in [this MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10457), `GHC.Parser` is not a good proxy for the root module of ghc-lib-parser anymore. fix this by precisely defining root modules per flavor and driving ghc-lib-parser module calculations from them.


shayne-fletcher:
cc @garyverhaegen-da @akrmn @dylant-da

garyverhaegen-da (approved):

PR #466, opened 2023-07-10
reuse ghc-lib-parser module definition lists in ghc-lib-test-mini-hlint

shayne-fletcher:
https://github.com/digital-asset/ghc-lib/pull/459 improved module selection for ghc-lib-parser. this change reuses the mechanism in ghc-lib-test-mini-hlint.

PR #467, opened 2023-07-15
ghc-lib module definition lists

shayne-fletcher:
use ghc-lib module definition lists & ghc make to calculate ghc-lib modules in the same manner we do as for ghc-lib-parser (#459, #466). we expect there to be no observable effect but its more principled than what it replaces and makes very explicit precisely how we define the ghc-lib-parser and ghc-lib apis per flavor.


shayne-fletcher:
cc @garyverhaegen-da @dylant-da @akrmn @samuel-williams-da

dylant-da (approved):

PR #486, opened 2023-09-24
bump some deps

shayne-fletcher:
almost more effort than it was worth but anyway, managed to bump a few packages while also removing `allow-newer` from hadrian's stack.yaml.


shayne-fletcher:
cc @garyverhaegen-da @akrmn @samuel-williams-da @dylant-da

PR #496, opened 2023-11-23
upgrade some packages

shayne-fletcher:
upgrade some packages


basvangijzel-DA (approved):
Thanks Shayne!

shayne-fletcher:
i took advantage of the fact this didn't land friday to sneak an extra change in. sorry about that @basvangijzel-DA.. will need another `azp /run`.

PR #501, opened 2024-01-19
ghc sources now assume GHC2021 extensions are in effect

shayne-fletcher:
- default language GHC2021 comes in this MR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11743
  - 9.4 and 9.6 don't understand GHC2021 as a language but do support -XGHC2021 so use that
- also required to patch `Safe` out of template haskell modules (not immediately obvious why that has come up just now)


garyverhaegen-da (approved):
Thank you!

Issue #483, opened 2023-08-30
`ghc-lib-parser` should depend on `happy >= 1.20`


michaelpj:
See https://github.com/haskell/haskell-language-server/issues/3783#issuecomment-1698715597

michaelpj:
Comparing with `lib:ghc`, both the 9.4 and 9.6 series depend on `happy > 1.20`

shayne-fletcher:
sorry i missed this @michaelpj. getting it fixed up now.

michaelpj:
The issue we had seemed to go away so I'm not 10% sure my diagnosis is correct, here!

PR #509, opened 2024-03-25
raise happy lower bound

shayne-fletcher:
(1) remove the workaround for alex breakage from https://github.com/digital-asset/ghc-lib/pull/508 (since https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12297 has landed)
(2) tighten lower bound on happy (fixes https://github.com/digital-asset/ghc-lib/issues/483 )


garyverhaegen-da (approved):
Thank you!

PR #514, opened 2024-05-02
add ghc-option -fno-safe-haskell to cabal files

shayne-fletcher:
[this refactor](https://gitlab.haskell.org/ghc/ghc/-/commit/4d78c53c527af05bc1b4944219fa88306449bae0) [broke ghc-lib-parser](https://github.com/shayne-fletcher/hlint-from-scratch/actions/runs/8916516869). the problem is we have been removing `Safe` from certain modules via code-mod. a better, less fragile approach is to to pass `-fno-safe-haskell` in the cabal ghc-options to ghc-lib-parser and ghc-lib which is what this PR does.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -76,7 +76,7 @@ data DaFlavor = DaFlavor
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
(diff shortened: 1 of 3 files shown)


garyverhaegen-da (approved):
Thank you!

PR #536, opened 2024-06-30
unix path config fix

shayne-fletcher:
fix a very obvious mistake


samuel-williams-da (approved):
Thanks for spotting!

samuel-williams-da:
This would have been caught by running github actions, we intend to discuss this internally soon, but we've been quite busy as of late.

shayne-fletcher:
@dylant-da thank-you 🙏

PR #537, opened 2024-07-01
ghc-9.6.6 support

shayne-fletcher:
support for ghc-9.6.6 released 2024-07-01.


ghc-lib-gen/src/GhclibgenFlavor.hs
```diff
@@ -42,7 +42,7 @@ ghcSeries = \case
     f | Ghc901 <= f && f < Ghc921 -> GHC_9_0
     f | Ghc921 <= f && f < Ghc941 -> GHC_9_2
     f | Ghc941 <= f && f < Ghc961 -> GHC_9_4
-    f | Ghc961 <= f && f < Ghc982 -> GHC_9_6
+    f | Ghc961 <= f && f < Ghc981 -> GHC_9_6
```
dylant-da, line 45: Is this right? Why was it 9.8.2 originally?

dylant-da (approved):
Aside from my question, looks good - approved, can merge - do you feel that my question requires any follow up?

ghc-lib-gen/src/GhclibgenFlavor.hs
shayne-fletcher: it's a good question. my view is that it was a mistake and clearly wrong in that it would have classified ghc-9.8.1 as a ghc 9.6 series compiler.

shayne-fletcher:
> Aside from my question, looks good - approved, can merge - do you feel that my question requires any follow up?

@dylant-da  sorry missed your question! have addressed it now and trust it satisfies!

dylant-da:
Thanks :) I think that's all your PRs sorted for now

Issue #534, opened 2024-06-25
GitHub actions disabled?


shayne-fletcher:
there are a couple of github workflows in the repo but they don't run on push. is this international? some config disabling actions?

@garyverhaegen-da

basvangijzel-DA:
@dylant-da FYI

shayne-fletcher:
with https://github.com/digital-asset/ghc-lib/pull/539 there is now a complete set of builds via github actions available such that if enabled, azure config can be removed and the project would be migrated over.

shayne-fletcher:
also worth noting that github builds take 50% of the azure build time.

paulbrauner-da:
Hi Shayne. Yes, github actions are disabled. We're looking into enabling them but in the meantime we have some questions on #540.

shayne-fletcher:
in https://github.com/digital-asset/ghc-lib/pull/540#issuecomment-2228931663 we talk about how we don't have a `--da` test (da-ghc-8.8.1) workflow. 

if we make the jump to github now as i advocate we'll have to work out how much we care about not having i suspect. personally, i'm ok with it linux + windows for that build is convincing enough to me and if there were going to be problems, with macos DA testing downstream would show it. if on other hand we do care enough we could keep the azure config for this build as long as intel x86_64 machines are provisioned that way there

paulbrauner-da:
Github actions are now enabled, see https://github.com/digital-asset/ghc-lib/pull/540#issuecomment-2233299172.

shayne-fletcher:
not quite. you can see from https://github.com/shayne-fletcher/ghc-lib/actions/runs/9992917978 for example that @nycnewman 's [c8ce1c7](https://github.com/shayne-fletcher/ghc-lib/commit/c8ce1c71d5b9e44a1ddb0e252eb0c5d65a63851d) caused actions to run on my fork but there still appears to be no jobs running on the digital asset repo itself.

shayne-fletcher:
> in [#540 (comment)](https://github.com/digital-asset/ghc-lib/pull/540#issuecomment-2228931663) we talk about how we don't have a `--da` test (da-ghc-8.8.1) workflow.
> 
> if we make the jump to github now as i advocate we'll have to work out how much we care about not having that build signal.
> 
> personally, i'm ok with it. linux + windows for that build is convincing enough for me. if there were going to be problems for DA, downstream testing would quickly show it. if on other hand we do care enough we could just keep the azure config for this one build.
> 
[...]

PR #540, opened 2024-07-15
less stack more cabal

shayne-fletcher:
this PR removes any and all uses of stack in this project. it is replaced completely by ghcup and cabal. there's a bit of cleanup work that will follow but the key objective of removing stack is done.


basvangijzel-DA:
Hey Shayne, thank you for the patience, we had some big deadlines last week!

paulbrauner-da:
Thanks Shayne! I have two questions:
- Would you mind summarizing the changes you made to the azure config outside of the stack->cabal migration? In particular, cases like `8.8.1-8.8.1` seem to have disappeared but `8.8.1-8.4.3` have appeared.
- Is there a reason why the github actions don't test exactly the same cases?
  - `8.8.1-8.4.3` doesn't seem to be tested, instead `8.4.3-8.4.3` is
  - `8.4.3` is not tested on macos.

shayne-fletcher:
@paulbrauner-da 

> Thanks Shayne! I have two questions:
> 
> * Would you mind summarizing the changes you made to the azure config outside of the stack->cabal migration? In particular, cases like `8.8.1-8.8.1` seem to have disappeared but `8.8.1-8.4.3` have appeared.
> * Is there a reason why the github actions don't test exactly the same cases?

no, no deep reason. i'll explain further below.
   
>   * `8.8.1-8.4.3` doesn't seem to be tested, instead `8.4.3-8.4.3` is

sure. it's not 8.4.3-8.4.3 being tested (is there a label wrong somewhere?) but ghc-flavor `--da` using 8.4.3 as a build compiler. what's important here is that we test `--da` (which is 8.8.1 GHC source customized by DA. the choices for such a test are build compilers <= 8.8.1. i happened to test with build compiler 8.4.3. if you think it important (i don't) we can go back to 8.8.1 for the build compiler for this test. let me know.

>   * `8.4.3` is not tested on macos.

actually, you'll see that it is tested on macos in the azure config. what you are referring to i think is that there isn't a github action for testing it. the reason for that is github actions appear to provide aarch64 machines for macos and ghcup fails to install these older compilers (8..4.3 and 8.8.1) on that architecture. on the other hand, azure appears to provide x86_64 and that's why that test remains in azure but is absent from the github actions.

shayne-fletcher:
@paulbrauner-da ping?

dylant-da:
Sounds good to me, but I'll leave it to Paul - I've pinged him on DA's Slack

basvangijzel-DA:
Hey Shayne, sorry for the delays, but we need to have a discussion with security internally to make sure changing GitHub actions, etc. is okay from their end.

shayne-fletcher:
> Hey Shayne, sorry for the delays, but we need to have a discussion with security internally to make sure changing GitHub actions, etc. is okay from their end.

that's fine, how we resolve https://github.com/digital-asset/ghc-lib/issues/534 is independent of this PR and the one that follow. nothing in these PRs stops testing using azure resources. blocking on this issue doesn't make any sense here.

paulbrauner-da:
Thank you for your patience @shayne-fletcher, we've now enabled GitHub actions, and I've approved the PR.

nycnewman:
I have enabled Github Actions for this one repo

shayne-fletcher:
> Thank you for your patience @shayne-fletcher, we've now enabled GitHub actions, and I've approved the PR.

thanks @paulbrauner-da  but please understand approving it isn't sufficient. i do not have write permission. you need to land it.

paulbrauner-da:
> thanks @paulbrauner-da but please understand approving it isn't sufficient. i do not have write permission. you need to land it.

My bad, done.

shayne-fletcher:
>   * `8.4.3` is not tested on macos.

@paulbrauner-da @dylant-da it looks like i've found a solution to this. there are some x86_64 runners available (https://github.com/actions/runner-images?tab=readme-ov-file#available-images). i'll circle back with a diff when the issues with the repo config that are currently blocking landing PRs are resolved.

PR #543, opened 2024-07-22
module dependency analysis improvements

shayne-fletcher:
simplify ghc -M and add new tests (in total now 6 x 3 builds).


shayne-fletcher:
@samuel-williams-da phew! that was a marathon 😄 this is finally ready to go. the last of the big changes.

samuel-williams-da:
@nycnewman Since this adds new jobs, we'll want to add these also to the required checks

samuel-williams-da:
Thanks a bunch for all this work @shayne-fletcher! I'll take a look tomorrow morning :)

samuel-williams-da:
A quick question of curiosity - I see you're now testing 9.0.2, what metric are you using to decide which versions need explicit tests?

shayne-fletcher:
> A quick question of curiosity - I see you're now testing 9.0.2, what metric are you using to decide which versions need explicit tests?

i'm working towards having a test of each series: 8.8, 8.10, 9.0, 9.2, 9.4., 9.6, 9.8, 9.10. according to that metric we are currently only lacking tests for 9.4 and 9.6 i think.

samuel-williams-da:
Ah right okay, good to know - thanks!
I wonder if GithubActions allows for a "parent" job that only passes if the rest do, this ends up being a lot of checks in the list that a repo admin (of which I am not) needs to keep up to date

shayne-fletcher:
> Ah right okay, good to know - thanks! I wonder if GithubActions allows for a "parent" job that only passes if the rest do, this ends up being a lot of checks in the list that a repo admin (of which I am not) needs to keep up to date

i expect now this volley of work is out of the way this list will be static and only extended on new ghc releases (next being ghc-12) so once a year? on the other hand, we could add that cost of some complexity fold all of these jobs into a single worflow file if that makes any difference? in short, i'm open to rephrasing the workflows to simplify the repo config mgt if that helps.

samuel-williams-da:
If GA doesn't have something simple for this, I'd rather leave as is, I'd still like to easily know which part has failed.
This isn't that deep, just a passing thought, not a concern for this PR.

shayne-fletcher:
> Thanks a bunch for all this work @shayne-fletcher! I'll take a look tomorrow morning :)

morning @samuel-williams-da ! how's it looking? ok to move forward on this one?

ghc-lib-gen.cabal
```diff
@@ -21,9 +21,9 @@ flag semaphore-compat
 common base
   default-language: Haskell2010
   ghc-options:
-    -Wall -Wincomplete-record-updates
-    -Wredundant-constraints -Widentities
-    -Wno-simplifiable-class-constraints
+  ghc-options:
```
samuel-williams-da, line 24:
```suggestion
```

ghc-lib-gen/src/Ghclibgen.hs
samuel-williams-da, line 1: Unsure why CI has repeated these warnings so many times, but it seems `System.Exit` and `System.Info.Extra` are unnecessary, same with `ghcApi` on line 1042. It seems Github Actions is showing warnings we hadn't noticed before.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -283,127 +284,101 @@ ghcLibExtraFiles ghcFlavor =
...
+        files <- filter ((`elem` [".hs-boot"]) . takeExtension) <$> listFilesRecursive path
+        forM_ files $ \file -> do
+          let p = fromJust (stripPrefix path file)
+              new_p = placeholderModulesDir </> p
+              dir = System.FilePath.takeDirectory new_p
+          createDirectoryIfMissing True dir
+          copyFile file new_p
+
+calcModuleDeps :: [FilePath] -> [FilePath] -> [FilePath] -> GhcFlavor -> FilePath -> String -> String
+calcModuleDeps includeDirs _hsSrcDirs hsSrcIncludes ghcFlavor cabalPackageDb ghcMakeModeOutputFile =
```
samuel-williams-da, line 337: I see quite a few options got removed here, why are they not needed anymore?

samuel-williams-da (approved):
I'm not entirely sure how you come to the changes needed for the variouus `Main.hs` files, but they seem fine. It would be helpful to arrange a walk through at some point so we can be on the same page as you when reviewing this code, we may reach out more formally about this after some discussion. (cc: @dylant-da)

shayne-fletcher:
> I'm not entirely sure how you come to the changes needed for the variouus `Main.hs` files, but they seem fine. It would be helpful to arrange a walk through at some point so we can be on the same page as you when reviewing this code, we may reach out more formally about this after some discussion. (cc: @dylant-da)

these files are just getting their import lists ordered alphabetically

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: yes, fixing this will follow in cleanup PRs

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: this is the exciting part for me. the fact is that over the course of these last 5 years through all the versions we've supported and the complexity of the procedure as conducted via stack we accumulated a lot of workarounds and hacks that we determined were necessary for the time. now with stack removed and reliance strictly on the cabal solver i've gone back and sought opportunities to remove these things. this is the result!

shayne-fletcher:
> > I'm not entirely sure how you come to the changes needed for the variouus `Main.hs` files, but they seem fine. It would be helpful to arrange a walk through at some point so we can be on the same page as you when reviewing this code, we may reach out more formally about this after some discussion. (cc: @dylant-da)
> 
> these files are just getting their import lists ordered alphabetically

i should be delighted to meet with you and walk through the role and meaning of these files and indeed anything else!

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: to be more explicit about what we've removed exactly here though: the preprocessor constants `STAGE=2`, `GHC_IN_GHCI` don't influence the outcome of the `ghc -M`. they probably never did and have always been redundant. the constant `GHCI` determines if `ghc -M` includes `GHCi.Run` and since the absence of that module does not affect the behavior of `ghc-lib-mini-compile-test` i judge it ok to exclude that module and drop the constant (less code is better).  the `ignore-package` directives were found to be unnecessary across all combinations tested so far (!) and so could be removed, similarly the `-package exceptions` has yet to be proven to be needed there (perhaps there's a config we haven't encountered yet that will require adding it back - we'll see). the `semaphoreCompatBootExists`  and `ghcPackagePath` related code got refactored into a better form using the `clear-package-db`, ..., `package-db` flags.

shayne-fletcher:
> > I'm not entirely sure how you come to the changes needed for the variouus `Main.hs` files, but they seem fine. It would be helpful to arrange a walk through at some point so we can be on the same page as you when reviewing this code, we may reach out more formally about this after some discussion. (cc: @dylant-da)
> 
> these files are just getting their import lists ordered alphabetically

in a nutshell, these `Main.hs` files *define* the ghc-lib-parser-api for a given ghc API (8.8 vs 9.0 vs. ... 9.10) etc. they are the minimal and complete set of GHC source modules required to compile the program `ghc-lib-test-mini-hlint` see around here https://github.com/shayne-fletcher/ghc-lib/blob/73e1ea48722ae23eeb138b18461adb4a1ab261db/examples/ghc-lib-test-mini-hlint/src/Main.hs#L15 for one place where use is made of them.

shayne-fletcher:
@samuel-williams-da i know these reviews have been hard work. i just want to say thanks  and for you to know that i appreciate you.

shayne-fletcher:
@samuel-williams-da @dylant-da might this now get merged please?

ghc-lib-gen/src/Ghclibgen.hs
samuel-williams-da: Oh nice! That is exciting! Happy to see simplifications here :)

samuel-williams-da:
Apologies, Friday was a company holiday - thanks for this work! I'll merge now

PR #546, opened 2024-07-25
remove stack yaml files

shayne-fletcher:


samuel-williams-da:
I assume this depends on the previous PR that moves away from Stack?

shayne-fletcher:
i think this PR can land before or after https://github.com/digital-asset/ghc-lib/pull/543. the current state of master doesn't make use of stack (since https://github.com/digital-asset/ghc-lib/pull/540).

samuel-williams-da (approved):

PR #549, opened 2024-07-30
Format with ormolu

shayne-fletcher:
prefer `join` over `++` and format with ormolu.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -63,7 +63,7 @@ data DaFlavor = DaFlavor
 
 -- Last tested gitlab.haskell.org/ghc/ghc.git at
 current :: String
-current = "12d3b66cedd3c80e7c1e030238c92d26631cab8d" -- 2024-07-17
+current = "23f50640e705c132f1a0689d4850866d0f0d76a6" -- 2024-07-29
 
 ghcFlavorOpt :: GhcFlavor -> String
 ghcFlavorOpt = \case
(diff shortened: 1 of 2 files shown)


samuel-williams-da:
Could we have this checked in CI/precommit of some kind?

samuel-williams-da:
Also, why the preference of `join` over `++` or `<>`?

shayne-fletcher:
> Could we have this checked in CI/precommit of some kind?

it sounds like a good idea! i'm not quite sure what you mean. could you elaborate?

shayne-fletcher:
> Also, why the preference of `join` over `++` or `<>`, at least in cases with only 2 lists?

ah! the end goal was to format with ormolu but they way it rendered (longer lists) with `++` just looked so terrible (to me) so i resolved to use `join` before applying it. i'm not opposed to `++` in general. i think the convention for `join` is working well here; if you see `++` now it's in the interior of a list, not top-level. i just blanket replaced the `++`s so didn't distinguish between short or long lists.

shayne-fletcher:
> Could we have this checked in CI/precommit of some kind?

maybe you mean, a CI job that builds ghc-lib from the current GHC HEAD commit? we can do that. it's implemented as `cabal run exe:ghc-lib-build-tool -- --ghc-flavor ""`

paulbrauner-da:
> > Could we have this checked in CI/precommit of some kind?
> 
> it sounds like a good idea! i'm not quite sure what you mean. could you elaborate?

Hi Shayne. I think what @samuel-williams-da means is we could have a pre-commit hook that runs the formatter, and we could have a CI tasks that fails if the code is not formatted.

samuel-williams-da:
Paul is correct, it should be relatively easy to add as a job.
Regarding `join`, I see ormolu does make it kind of ugly, I don't mind either way

shayne-fletcher:
> Paul is correct, it should be relatively easy to add as a job. 

yes, it's easy. similarly we should reintroduce hlint. can these things be done in a later PR though please?

samuel-williams-da:
Sure I'm okay with that

PR #550, opened 2024-08-01
ormolu check

shayne-fletcher:
add an ormolu check workflow covering CI.hs, ghc-lib-gen. required formatting some more source files to get the test green.

### DIFF
--- a/.github/workflows/ormolu-check.yml
+++ b/.github/workflows/ormolu-check.yml
@@ -1,0 +1,16 @@
+name: ormolu-check
+on:
+  push:
+  pull_request:
+jobs:
+  ormolu:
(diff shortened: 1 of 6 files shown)


samuel-williams-da:
Any chance we could get this on a git pre-commit hook? If not, at least make running the formatter easy (perhaps a shell file that passes in the patterns)

shayne-fletcher:
i think that's somewhat orthogonal to what this pr offers? i don't have much interest in pursuing this aspect further right now. perhaps you (or someone else) could look into adding that?

samuel-williams-da (approved):

samuel-williams-da:
Sure! I'll take a look next week :)

PR #551, opened 2024-08-02
add hlint check workflow

shayne-fletcher:
a workflow that checks for hlint warnings


dylant-da (approved):
Sorry about the slow review, it's been a 4-day weekend here in the UK
LGTM

PR #552, opened 2024-08-02
add a cron job for master building from GHC HEAD

shayne-fletcher:
this workflow will give us an automated signal when a commit in GHC upstream breaks ghc-lib. the job it defines runs daily at 12:00am. it runs the ghc-lib CI process using the prevailing GHC HEAD commit.


samuel-williams-da:
This is nice, thanks Shayne!

samuel-williams-da (approved):

samuel-williams-da:
cc: @nycnewman as this may affect CI cost?

Issue #563, opened 2024-10-22
ghc-lib-parser-9.8.2.20240223 failed to build with ghc-9.8.3


juhp:
For Stackage Nightly with ghc-9.8.3:

```
       Building library for ghc-lib-parser-9.8.2.20240223..                                                                         
       [155 of 433] Compiling GHC.Unit.Types                                                                                        
                                                                                                                                    
       /home/curators/work/unpack-dir/unpacked/ghc-lib-parser-9.8.2.20240223-2c6194cfab3c6d198ba355d48ffc919705e56dc74190a757398fed9
667763406/compiler/GHC/Unit/Types.hs:127:25: error: [GHC-87543]                                                                     
           Ambiguous occurrence ‘Unit’.                                                                                             
           It could refer to                                                                                                        
              either ‘Control.DeepSeq.Unit’,                                                                                        
                     imported from ‘Control.DeepSeq’ at compiler/GHC/Unit/Types.hs:104:1-22,                                        
                  or ‘GHC.Unit.Types.Unit’,                                                                                         
                     defined at compiler/GHC/Unit/Types.hs:289:1.                                                                   
           |                                                                                                                        
       127 | type Module = GenModule Unit                                                                                           
           |                         ^^^^                                                                                           
                                                                                                                                    
       /home/curators/work/unpack-dir/unpacked/ghc-lib-parser-9.8.2.20240223-2c6194cfab3c6d198ba355d48ffc919705e56dc74190a757398fed9
667763406/compiler/GHC/Unit/Types.hs:334:14: error: [GHC-87543]                                                                     
           Ambiguous occurrence ‘Unit’.                                                                                             
           It could refer to                                                                                                        
              either ‘Control.DeepSeq.Unit’,                                                                                        
                     imported from ‘Control.DeepSeq’ at compiler/GHC/Unit/Types.hs:104:1-22,                                        
                  or ‘GHC.Unit.Types.Unit’,                                                                                         
                     defined at compiler/GHC/Unit/Types.hs:289:1.
           |
       334 | instance Ord Unit where
           |              ^^^^
        
       /home/curators/work/unpack-dir/unpacked/ghc-lib-parser-9.8.2.20240223-2c6194cfab3c6d198ba355d48ffc919705e56dc74190a757398fed9
667763406/compiler/GHC/Unit/Types.hs:337:15: error: [GHC-87543]
           Ambiguous occurrence ‘Unit’.
           It could refer to
[...]

PR #565, opened 2024-10-23
fix happy bounds

shayne-fletcher:
motivation for this change discussed in https://github.com/digital-asset/ghc-lib/issues/563. for >=9.8 flavors, allow happy-2.0.2 in ghc-lib-parser build-tool-depends.

m4/fptools_happy.m4 informs configure which as of https://gitlab.haskell.org/ghc/ghc/-/commit/0029ca91c845dd4530eb2c4606ad5bd59775cec2 allows ==1.20.* || >= 2.0.2 && < 2.1

### DIFF
--- a/ghc-lib-gen.cabal
+++ b/ghc-lib-gen.cabal
@@ -52,7 +52,10 @@ executable ghc-lib-gen
 
 executable ghc-lib-build-tool
   import: base
-  build-tool-depends: alex:alex, happy:happy < 2.0
+  if impl (ghc > 9.12.0)
+    build-tool-depends: alex:alex, happy:happy == 1.20.* || >= 2.0.2 && < 2.1
+  else
+    build-tool-depends: alex:alex, happy:happy < 2.0
   build-depends:
     directory, filepath, time, extra, optparse-applicative
   if flag(semaphore-compat)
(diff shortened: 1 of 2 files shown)


ghc-lib-gen.cabal
```diff
@@ -52,7 +52,10 @@ executable ghc-lib-gen
 
 executable ghc-lib-build-tool
   import: base
-  build-tool-depends: alex:alex, happy:happy < 2.0
+  if impl (ghc > 9.12.0)
```
dylant-da, line 55: Should this bound instead check for 9.8?

dylant-da (approved):
Pending comment, approved

ghc-lib-gen.cabal
shayne-fletcher: no, it's a bit subtle. this choice affects the happy that we'll use in ghc-lib-gen that invokes the GHC configure stage in order to compute the modules that go into each of ghc-lib-parser and ghc-lib, not the happy build-tool-depends of those packages (building these packages does not involve configure and generally don't fail with happy-2.0.2, i.e. the point raised in #563 ). except on the current HEAD of the GHC source tree, configure will fail with happy > 1.20 (the change that admits happy 2.0.2 only went on master 2 weeks ago {see the referenced commit above}). so this checks that the GHC source we are working with is at least > ghc-9.12.1 (equivalently, the minimum ghc-lib-parser flavor is ghc-9.12.1) before enabling 2.0.2 for happy for the purposes of ghc-lib-gen. in all other cases, 1.20.* works and will pass configure just fine so in tl;dr, this is what we want.

PR #568, opened 2024-11-03
workaround ghc-9.10.1/deepseq ambiguity

shayne-fletcher:
(1) ghc-9.8.3 build compiler ships with deepseq 15.1.0 which breaks building ghc-9.10.1 source (showing the symptoms of https://github.com/digital-asset/ghc-lib/issues/563 see https://github.com/digital-asset/ghc-lib/issues/563#issuecomment-2453509242). the fix is already on ghc-9.12 and HEAD. the fix here is to patch ghc-lib-parser flavor ghc-9.10.1 and re-release (ghc-lib-9.10.1.20241103). i updated the happy bounds manually to `happy:happy == 1.20.* || == 2.0.2 || >= 2.1.2 && < 2.2` since there's alread a diff in flight for that in https://github.com/digital-asset/ghc-lib/pull/567.

(2) it looks like flavor ghc-9.8.3 was released from the ghc-9.8 branch, not the ghc-9.8.3 branch (my fault i guess) so that's fixed here as well. i've pushed a new release for that (ghc-lib-9.8.3.20241103) flavor and manually adjusted the happy bounds for that too `happy:happy == 1.20.* || == 2.0.2 || >= 2.1.2 && < 2.2`.

### DIFF
--- a/CI.hs
+++ b/CI.hs
@@ -496,7 +496,7 @@ buildDists ghcFlavor noGhcCheckout noBuilds versionSuffix = do
     branch = \case
       Ghc9121 -> "ghc-9.12"
       Ghc9101 -> "ghc-9.10.1-release"
-      Ghc983 -> "ghc-9.8"
+      Ghc983 -> "ghc-9.8.3-release"
       Ghc982 -> "ghc-9.8.2-release"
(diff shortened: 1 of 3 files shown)


dylant-da (approved):
LGTM

PR #570, opened 2024-11-26
Remove pSupportedExts from ParserOpts

shayne-fletcher:
adapt to `caaf53881d5cc82ebff617f39ad5363429d2eccf`.

two small changes:
   - new argument to to `getOptions` 
   - supply definition for `sUnitSettings` of `Settings`.

### DIFF
--- a/examples/ghc-lib-test-mini-compile/src/Main.hs
+++ b/examples/ghc-lib-test-mini-compile/src/Main.hs
@@ -193,6 +193,23 @@ mkDynFlags filename s = do
       (dflags, _, _) <- parseDynamicFilePragma dflags0 opts
       return dflags
 
+#elif (defined (GHC_9_6) || defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12))
+
+  let baseFlags =
+        (defaultDynFlags fakeSettings) {
+          ghcLink = NoLink
+        , backend = noBackend
+        , homeUnitId_ = toUnitId (stringToUnit ghclibPrimUnitId)
+        }
+  parsePragmasIntoDynFlags filename s baseFlags
+  where
+    parsePragmasIntoDynFlags :: String -> String -> DynFlags -> IO DynFlags
+    parsePragmasIntoDynFlags filepath contents dflags0 = do
+      let (_, opts) = getOptions (initParserOpts dflags0)
+                        (stringToStringBuffer contents) filepath
+      (dflags, _, _) <- parseDynamicFilePragma dflags0 opts
+      return dflags
+
 #else
 
   let baseFlags =
@@ -206,6 +223,7 @@ mkDynFlags filename s = do
     parsePragmasIntoDynFlags :: String -> String -> DynFlags -> IO DynFlags
     parsePragmasIntoDynFlags filepath contents dflags0 = do
       let (_, opts) = getOptions (initParserOpts dflags0)
+                        (supportedLanguagePragmas dflags0)
                         (stringToStringBuffer contents) filepath
       (dflags, _, _) <- parseDynamicFilePragma dflags0 opts
       return dflags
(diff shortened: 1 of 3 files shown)


examples/ghc-lib-test-mini-compile/src/Main.hs
```diff
@@ -394,16 +412,51 @@ fakeSettings = Settings {
...
 
+#elif (defined (GHC_9_4) || defined (GHC_9_6) || defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12))
+    sGhcNameVersion=ghcNameVersion
+  , sFileSettings=fileSettings
+  , sTargetPlatform=platform
+  , sPlatformMisc=platformMisc
+  , sToolSettings=toolSettings
+  }
+  where
+    fileSettings = FileSettings {
```
dylant-da, line 423: Since the only difference between GHC_9_2 and GHC_9_4 - GHC_9_12 and then GHC_9_14 is the FileSettings struct, can we maybe use a nested if statement to only change fileSettings between the two versions? e.g.

```haskell
#else
    {- defined (GHC_9_4) || defined (GHC_9_6) || defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12) || defined (GHC_9_14) -}
    sGhcNameVersion=ghcNameVersion
    ...
    #if defined(GHC_9_14)
        sUnitSettings=unitSettings
    #endif
    fileSettings = FileSettings {
    #if defined (GHC_9_2)
        fileSettings_tmpDir="."
        ...
    #else
        fileSettings_topDir="."
        ...
    #endif
    
    toolSettings = ToolSettings {
    ...
#else
...
```

I see why this may look more confusing than just flattening all the cases, so it's up to you. The main advantage is that we only need to change a common line like `platformMisc = PlatformMisc {` in one place.

examples/ghc-lib-test-mini-hlint/src/Main.hs
```diff
@@ -181,11 +181,30 @@ parsePragmasIntoDynFlags flags filepath str =
...
+        sDoc : _ -> do putStrLn sDoc; return Nothing
+      where
+        sDocs = [ showSDoc flags msg | msg <- pprMsgEnvelopeBagWithLocDefault . getMessages $ srcErrorMessages msgs ]
 #else
-    {- defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12) || defined (GHC_9_14) -}
+    {- defined (GHC_9_14) -}
 
   catchErrors $ do
     let (_, opts) = getOptions (initParserOpts flags)
+                      (supportedLanguagePragmas flags)
```
dylant-da, line 207: Similar idea about nested preprocessor if statements here.

examples/ghc-lib-test-mini-compile/src/Main.hs
```diff
@@ -193,6 +193,23 @@ mkDynFlags filename s = do
       (dflags, _, _) <- parseDynamicFilePragma dflags0 opts
       return dflags
 
+#elif (defined (GHC_9_6) || defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12))
```
dylant-da, line 196: Similar idea for nested if statements here

examples/ghc-lib-test-mini-hlint/src/Main.hs
```diff
@@ -631,16 +650,46 @@ fakeSettings = Settings {
 
     platform=genericPlatform
 
+#elif (defined (GHC_9_4) || defined (GHC_9_6) || defined (GHC_9_8) || defined (GHC_9_10) || defined (GHC_9_12))
```
dylant-da, line 653: Similar idea for nested if statements here

dylant-da (approved):

shayne-fletcher:
i've been down this road before and the choices i've made here ("flattening the cases") is deliberate based on the experience even if it does mean more lines of code 🙂  if it's ok with you i'd prefer to lay it out this way

dylant-da:
I figured there was a reason, considering that I noticed just about everything is flattened in the rest of the codebase too. Thanks a lot for getting this through!

PR #577, opened 2024-12-31
restrict alex < 3.5.2.0

shayne-fletcher:
alex-3.5.2.0 was released to hackage on 2024-12-30. it produces errors like this
```
compiler/GHC/Parser/Lexer.x:3471:12: error: [GHC-31744]
    Duplicate INLINE pragmas for ‘alexScanUser’
    at /home/runner/work/ghc-lib/ghc-lib/dist-newstyle/build/x86_64-linux/ghc-9.10.1/ghc-lib-parser-0.20241231/build/GHC/Parser/Lexer.hs:1383:12-23
       compiler/GHC/Parser/Lexer.x:3471:12-23
     |
3471 | -- in this file, alexScanUser gets broken out into a separate function and
     |            ^^^^^^^^^^^^
```
on flavors >= ghc-9.12.1 e.g. https://github.com/digital-asset/ghc-lib/actions/runs/12553438884. verified that this is a problem for GHC too. see https://github.com/haskell/alex/issues/266.

incidentally, remove `extraCabalFlags` from CI.hs (they are needed any more) and slightly modify the happy bounds for ghc < 9.8 to workaround a build failure for workflow ghc-flavor 9.0.2 with build compiler 8.10.7 (i have no idea why that decided to break just now).

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -1314,7 +1314,7 @@ libBinParserLibModules ghcFlavor = do
 
 happyBounds :: GhcFlavor -> String
 happyBounds ghcFlavor
-  | series < GHC_9_8 = "== 1.20.*"
+  | series < GHC_9_8 = "< 1.21"
   | otherwise = "== 1.20.* || == 2.0.2 || >= 2.1.2 && < 2.2" -- c.f. m4/fptools_happy.m4
   where
     series = ghcSeries ghcFlavor
@@ -1393,7 +1393,7 @@ generateGhcLibCabal ghcFlavor customCppOpts = do
         "    build-depends:"
       ],
       indent2 (Data.List.NonEmpty.toList (withCommas (ghcLibBuildDepends ghcFlavor))),
-      ["    build-tool-depends: alex:alex >= 3.1, " ++ "happy:happy " ++ happyBounds ghcFlavor],
+      ["    build-tool-depends: alex:alex >= 3.1 && < 3.5.2.0, " ++ "happy:happy " ++ happyBounds ghcFlavor],
       ["    other-extensions:"],
       indent2 (askField lib "other-extensions:"),
       ["    default-extensions:"],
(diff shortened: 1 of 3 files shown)


ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -1563,20 +1575,12 @@ generatePrerequisites ghcFlavor = do
...
-  let hadrianExtraCabalFlags = "--allow-newer "
-#else
-  let hadrianExtraCabalFlags = ""
-#endif
-
   system_ "bash -c ./boot"
   system_ "bash -c \"./configure --enable-tarballs-autodownload\""
   withCurrentDirectory "hadrian" $ do
-    system_ $ "cabal build " ++ hadrianExtraCabalFlags ++ "exe:hadrian --ghc-options=-j"
+    system_ $ "cabal build exe:hadrian --ghc-options=-j"
```
dylant-da, line 1581: The warning suggests we drop this dollar, up to you

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -1314,7 +1326,7 @@ libBinParserLibModules ghcFlavor = do
 
 happyBounds :: GhcFlavor -> String
 happyBounds ghcFlavor
-  | series < GHC_9_8 = "== 1.20.*"
+  | series < GHC_9_8 = "< 1.21"
```
dylant-da, line 1329: What's the difference here? Is there a situation where we need to supply 1.19 to GHC9.8?

dylant-da (approved):
LGTM - minor nitpicks, but it's all up to you

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: i pushed a new commit for it.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: yes, ghc-lib-ghc-9.0.2-ghc-8.10.7.yml (i have no idea why this didn't get picked up before now).

PR #579, opened 2025-01-24
fix CPP guards around ghc_unique_counter64

shayne-fletcher:
[compiler: Fix CPP guards around ghc_unique_counter64](https://gitlab.haskell.org/ghc/ghc/-/commit/595013d41464c1e328369bb81ce0ea2814e91b68)


dylant-da (approved):
Sorry about the delay on this - approved

PR #580, opened 2025-01-27
ghc-prim is merged with ghc-internal

shayne-fletcher:
[this MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/13752) merges ghc-prim into ghc-internal.

in this PR i manage to get ghc-lib-parser working again. the main ideas are: for ghc-9.12.1 ghc -M does not need any source paths to ghc-internal modules and neither do any need mentioning in either of the generated cabal files. that's not true for ghc-9.10.1 unfortunately so put some cpp around that.

the build of ghc-lib/ghc-lib-test-mini-compile succeeds but i've disabled testing of ghc-lib-test-mini-compile. this is a failing testg since this merge and i don't know how to fix it. i tested an equivalent program with a build of ghc after the MR and it fails in the same way (can't resolve `GHC.Internal.Prim`). i raised an issue upstream https://gitlab.haskell.org/ghc/ghc/-/issues/25686

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -80,11 +80,17 @@ cabalFileLibraries ghcFlavor =
     ]
 
 -- C-preprocessor "include dirs" for 'ghc-lib-parser'.
-ghcLibParserIncludeDirs :: GhcFlavor -> [FilePath]
-ghcLibParserIncludeDirs ghcFlavor =
+ghcLibParserIncludeDirs :: Bool -> GhcFlavor -> [FilePath]
+ghcLibParserIncludeDirs _forDepends ghcFlavor =
   join
     [ case ghcSeries ghcFlavor of
-        series | series > GHC_9_10 -> ["libraries/ghc-internal/include", "rts/include", "rts/include/stg"]
+        series | series >= GHC_9_10 ->
+                 [ "rts/include"
+                 , "rts/include/stg"
+#if __GLASGOW_HASKELL__ == 910
+                , "libraries/ghc-internal/include"
+#endif
+                 ]
         series | series > GHC_9_8 -> ["rts/include", "rts/include/stg"]
         series | series >= GHC_9_4 -> ["rts/include"] -- ghcconfig.h, ghcversion.h
         series | series < GHC_9_4 -> ["includes"] -- ghcconfig.h, MachDeps.h, MachRegs.h, CodeGen.Platform.hs
@@ -95,7 +101,7 @@ ghcLibParserIncludeDirs ghcFlavor =
     ]
 
 -- C-preprocessor "include dirs" for 'ghc-lib'.
-ghcLibIncludeDirs :: GhcFlavor -> [FilePath]
+ghcLibIncludeDirs :: Bool -> GhcFlavor -> [FilePath]
 ghcLibIncludeDirs = ghcLibParserIncludeDirs
 
 -- Sort by length so the longest paths are at the front. We do this
@@ -109,14 +115,18 @@ sortDiffListByLength all excludes =
 -- The "hs-source-dirs" universe.
 allHsSrcDirs :: Bool -> GhcFlavor -> [Cabal] -> [FilePath]
 allHsSrcDirs forDepends ghcFlavor lib =
-  join
+  filter (/= "libraries/ghc-boot-th/../ghc-internal/src") (join
     [ [stage0Compiler],
+#if __GLASGOW_HASKELL__ == 910
+      -- ghc -M needs source paths to GHC.Internal modules
+      ["ghc-lib/stage0/libraries/ghc-internal/build" | forDepends],
+      [dir | forDepends, dir <- ["libraries/ghc-boot-th-internal", "libraries/ghc-internal/src"]],
+#endif
       [dir | forDepends, dir <- [stage0Ghci, stage0GhcHeap]],
       [stage0GhcBoot | ghcSeries ghcFlavor >= GHC_8_10],
       map takeDirectory (cabalFileLibraries ghcFlavor),
-      map (dropTrailingPathSeparator . normalise) (askFiles lib "hs-source-dirs:"),
-      ["libraries/ghc-boot-th-internal"]
-    ]
+      map (dropTrailingPathSeparator . normalise) (askFiles lib "hs-source-dirs:")
+    ])
 
 -- The "hs-source-dirs" for 'ghc-lib-parser'.
 ghcLibParserHsSrcDirs :: Bool -> GhcFlavor -> [Cabal] -> [FilePath]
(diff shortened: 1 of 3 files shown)


shayne-fletcher:
the latest on https://gitlab.haskell.org/ghc/ghc/-/issues/25686 is that we probably have a fix for the failing tes

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -80,11 +80,17 @@ cabalFileLibraries ghcFlavor =
     ]
 
 -- C-preprocessor "include dirs" for 'ghc-lib-parser'.
-ghcLibParserIncludeDirs :: GhcFlavor -> [FilePath]
-ghcLibParserIncludeDirs ghcFlavor =
+ghcLibParserIncludeDirs :: Bool -> GhcFlavor -> [FilePath]
+ghcLibParserIncludeDirs _forDepends ghcFlavor =
```
dylant-da, line 84: It looks like this _forDepends parameter is no longer used

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -95,7 +101,7 @@ ghcLibParserIncludeDirs ghcFlavor =
     ]
 
 -- C-preprocessor "include dirs" for 'ghc-lib'.
-ghcLibIncludeDirs :: GhcFlavor -> [FilePath]
+ghcLibIncludeDirs :: Bool -> GhcFlavor -> [FilePath]
 ghcLibIncludeDirs = ghcLibParserIncludeDirs
```
dylant-da, line 105: If this defers to ghcLibParserIncludeDirs, then this Bool argument is also unused, I think

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -365,7 +385,7 @@ calcParserModules ghcFlavor = do
   copyFile mainFile rootModulePath
   lib <- mapM readCabalFile (cabalFileLibraries ghcFlavor)
   cabalPackageDb <- cabalPackageDb =<< ghcNumericVersion
-  let includeDirs = map ("-I" ++) (ghcLibParserIncludeDirs ghcFlavor)
+  let includeDirs = map ("-I" ++) (ghcLibParserIncludeDirs True ghcFlavor)
```
dylant-da, line 388: Since ghcLibParserIncludeDirs ignores its first argument, we can drop this `True`

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -382,7 +402,7 @@ calcLibModules ghcFlavor = do
   copyFile mainFile rootModulePath
   lib <- mapM readCabalFile (cabalFileLibraries ghcFlavor)
   cabalPackageDb <- cabalPackageDb =<< ghcNumericVersion
-  let includeDirs = map ("-I" ++) (ghcLibIncludeDirs ghcFlavor)
+  let includeDirs = map ("-I" ++) (ghcLibIncludeDirs True ghcFlavor)
```
dylant-da, line 405: Since ghcLibIncludeDirs ignores its first argument, we can drop this `True`

Aside: Seeing as ghcLibIncludeDirs is the same ghcLibParserIncludeDirs now, is there a reason we keep them as separate identifiers?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -1353,8 +1383,8 @@ generateGhcLibCabal ghcFlavor customCppOpts = do
             (Set.fromList parserModules))
       ```
   -}
-  let hsSrcDirs = replace ["libraries/ghc-boot-th/../ghc-internal/src"] [] (ghcLibHsSrcDirs False ghcFlavor lib)
-  let includeDirs = replace ["libraries/ghc-internal/include"] [] (ghcLibIncludeDirs ghcFlavor)
+  let hsSrcDirs = ghcLibHsSrcDirs False ghcFlavor lib
+  let includeDirs = ghcLibIncludeDirs False ghcFlavor
```
dylant-da, line 1387: Same as above?

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -1467,10 +1498,8 @@ performExtraFilesSubstitutions ghcFlavor files =
 generateGhcLibParserCabal :: GhcFlavor -> [String] -> IO ()
 generateGhcLibParserCabal ghcFlavor customCppOpts = do
   (lib, _bin, parserModules, _) <- libBinParserLibModules ghcFlavor
-  -- Remove 'ghc-interal/src', 'ghc-boot-th/internal' etc. They are
-  -- dealt with later explicitly.
-  let hsSrcDirs = filter (not . isInfixOf "internal") $ ghcLibParserHsSrcDirs False ghcFlavor lib
-  let includeDirs = filter (/= "libraries/ghc-internal/include") $ ghcLibParserIncludeDirs ghcFlavor
+  let hsSrcDirs = ghcLibParserHsSrcDirs False ghcFlavor lib
+  let includeDirs = ghcLibParserIncludeDirs False ghcFlavor
```
dylant-da, line 1502: Same here

dylant-da (approved):
Looks good, thank you!

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: i introduced this generalizationresearching solutions to this PR. in the end we don't use it but it makes sense to leave it plumbed through now that it's there. chances are it will prove useful in the future.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: it is true that `forDepends` is ignored. it's ok that it should remain.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: here we pass through the value of `forDepends` from a context in which it is true. let us continue to do that though it's currently not strictly required.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: yes, a formality where in that one can reasonably be expected to differ from the other and at some point probably will . expressing them as one is the other is a shortcut.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: in this case we pass through `forDepends` in a context in which it is `False`. in `ghcLibIncludeDirs` it is not currently inspected but that's ok. it's good to have the capability there.

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: `forDepends` in a context in which the source path is not going to be put to use in ghc -M. currently ignored by `ghcLibParserIncludeDirs` but available.

shayne-fletcher:
> Looks good, thank you - minor points:

thanks for the careful review @dylant-da . i hope you'll be ok with the generalizations of `ghcLibParserIncludeDirs ` and `ghcLibIncludeDirs `. they are logically functions of `ghcFlavor` even though they are constant with respect to flavor at this time.

examples/ghc-lib-test-mini-compile/test/Main.hs
```diff
@@ -2,7 +2,7 @@
 -- its affiliates. All rights reserved. SPDX-License-Identifier:
 -- (Apache-2.0 OR BSD-3-Clause)
 
-{-# OPTIONS_GHC -Werror=unused-imports -Werror=unused-local-binds -Werror=unused-top-binds -Werror=orphans #-}
+-- {-# OPTIONS_GHC -Werror=unused-imports -Werror=unused-local-binds -Werror=unused-top-binds -Werror=orphans #-}
```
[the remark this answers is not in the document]
shayne-fletcher: i commented it out for debugging. good catch!

dylant-da:
Sure, that makes sense, I'm alright with that generalization.

shayne-fletcher:
> Sure, that makes sense, I'm alright with that generalization. Feel free to merge once CI passes.

I can't merge @dylant-da . I don't have write access.

dylant-da:
Sorry about that, I should've realized

PR #593, opened 2025-03-27
fetching specific commits

shayne-fletcher:
`git checkout -f SHA` can fail for out-of-tree commits so be sure to `git fetch SHA` first.


dylant-da:
This *may* need rebasing on top of https://github.com/digital-asset/ghc-lib/pull/592, we'll wait and see.

dylant-da (approved):

PR #596, opened 2025-04-17
fix multiple declarations of ghc_unique_counter64

### LINKED ISSUE #595: ghc-lib-parser 9.12.2 fails to build with ghc-9.8.4 and ghc-lib-parser-9.10.1 fails to build with ghc-9.6.7
Probably related to this https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10568
(patch here https://gitlab.haskell.org/ghc/ghc/-/commit/e421ec3c8fc9688b3f7ee08c0d1baaf1cb7aa750)
`Linking /root/github/stylish-haskell/dist-newstyle/build/x86_64-linux/ghc-9.8.4/stylish-haskell-0.14.6.0/t/stylish-haskell-tests/build/stylish-haskell-tests/stylish-haskell-tests
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: /usr/lib64/ghc-9.8.4/lib/../lib/x86_64-linux-ghc-9.8.4/rts-1.0.2/libHSrts-1.0.2.a(Globals.o):(.bss+0x0): multiple definition of `ghc_unique_counter64'; /root/.local/state/cabal/store/ghc-9.8.4/ghc-lib-parser-9.12.2.20250320-48b5863d60855e9cdc5daef7f57fb49458d3bd0c527dd8950e22b0a76fe73a30/lib/libHSghc-lib-parser-9.12.2.20250320-48b5863d60855e9cdc5daef7f57fb49458d3bd0c527dd8950e22b0a76fe73a30.a(genSym.o):(.bss+0x0): first defined here`

GuillaumedeVolpiano:
Fixes #595 .

So this is basically your patch. I removed the guard about ghcFlavor because we need it for Ghc9122 too.

One thing: unless I'm mistaken, this mike cause issues for people using ghc-9.8.0 to 9.8.3 as the offending patch has only been backported to ghc-9.8.4. If this is something you care about, then we would need something like

. replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,9,0,0)" "#if !MIN_VERSION_GLASGOW_HASKELL(9,6,7,0) || !MIN_VERSION_GLASGOW_HASKELL(9,8,0) && !MIN_VERSION_GLASGOW_HASKELL(9,8,4,0)"

What do you think?

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -963,11 +963,15 @@ mangleCSymbols ghcFlavor = do
         . prefixSymbol genSym
         . prefixSymbol initGenSym
           =<< readFile' file
-  when (ghcFlavor <= Ghc9121) $
-    let file = "compiler/cbits/genSym.c"
-    in writeFile file
-       . replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,8,4,0)" "#if !MIN_VERSION_GLASGOW_HASKELL(9,6,7,0)"
-       =<< readFile' file
+  let file = "compiler/cbits/genSym.c"
+      replacement = "#if !MIN_VERSION_GLASGOW_HASKELL(9,6,7,0)\n"
+                        <> "HsWord64 ghc_unique_counter64 = 0;\n"
+                        <> "#elif MIN_VERSION_GLASGOW_HASKELL(9,8,0,0) "
+                        <> "&& !MIN_VERSION_GLASGOW_HASKELL(9,8,4,0)"
+   in writeFile file
+        . replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,8,4,0)" replacement
+        . replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,9,0,0)" replacement
+        =<< readFile' file
   when (ghcFlavor == Ghc984) $
     let file = "compiler/cbits/genSym.c"
      in writeFile file


shayne-fletcher:
> Fixes #595 .
> 
> So this is basically your patch. I removed the guard about ghcFlavor because we need it for Ghc9122 too.
> 
> One thing: unless I'm mistaken, this mike cause issues for people using ghc-9.8.0 to 9.8.3 as the offending patch has only been backported to ghc-9.8.4. If this is something you care about, then we would need something like
> 
> . replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,9,0,0)" "#if !MIN_VERSION_GLASGOW_HASKELL(9,6,7,0) || !MIN_VERSION_GLASGOW_HASKELL(9,8,0) && !MIN_VERSION_GLASGOW_HASKELL(9,8,4,0)"
> 
> What do you think?

we care very much that flavor 9.10.1 work with [9.6.0, 9.10.1] and that flavor 9.12.2 work with [9.8.0, 9.12.2]. any chance we can prevail on you to run some tests with for those you've identified as worrisome?

GuillaumedeVolpiano:
- [ ] ghc-lib-parser-9.12.2
  - [ ] ghc-9.8.3
  - [ ] ghc-9.8.2
  - [ ] ghc-9.8.1
- [ ] ghc-lib-parser-9.10.1
  - [ ] ghc-9.8.3
  - [ ] ghc-9.8.2
  - [ ] ghc-9.8.1
  - [ ] ghc-9.6.6
  - [ ] ghc-9.6.5
  - [ ] ghc-9.6.4
  - [ ] ghc-9.6.3
  - [ ] ghc-9.6.2
  - [ ] ghc-9.6.1

GuillaumedeVolpiano:
Just realised there was the same issue with [ghc-9.8.1-ghc-9.8.3] with ghc-lib-parser-9.8.5 (your other replace). Some more tests to run.

GuillaumedeVolpiano:
@shayne-fletcher :

I wonder if the
  `. replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,8,4)"`
line is necessary anymore ? It only makes sense for the ghc-9.8.4 tag of ghc but when I try to build —ghc-flavor=ghc-9.8.4, it actually builds ghc-9.8.5 ("current" ghc-9.8.4) which is patched and doesn’t need (nor uses) the changes.

that said, ready for review

shayne-fletcher:
> @shayne-fletcher :
> 
> I wonder if the `. replace "#if !MIN_VERSION_GLASGOW_HASKELL(9,8,4)"` line is necessary anymore ? It only makes sense for the ghc-9.8.4 tag of ghc but when I try to build —ghc-flavor=ghc-9.8.4, it actually builds ghc-9.8.5 ("current" ghc-9.8.4) which is patched and doesn’t need (nor uses) the changes.
> 
> that said, ready for review

i think it's strictly inclusive if we leave it in right? that is it shouldn't matter if stays?

i would say perhaps let's not include these format changes unintentionally. mind reverting that for this PR?

GuillaumedeVolpiano:
No it doesn't, which is why I left it.

Sorry for the format changes, they should be reverted now.

GuillaumedeVolpiano:
Exactly

GuillaumedeVolpiano:
We might want to hold on until we can test ghc-9.10.2 ? It's not yet available through ghcup and I won't be able to do a full build this weekend, but I should be able to check early next week.

shayne-fletcher:
> Exactly, apart for the lines of comments, I didn't put them. Want them in?

not necessary

dylant-da (approved):
Looks good, sorry for the late review / approval.

PR #599, opened 2025-05-03
new flavor ghc-9.10.2

shayne-fletcher:
uploaded to hackage today

### DIFF
--- a/ghc-lib-gen/src/Ghclibgen.hs
+++ b/ghc-lib-gen/src/Ghclibgen.hs
@@ -84,10 +84,17 @@ ghcLibParserIncludeDirs :: Bool -> GhcFlavor -> [FilePath]
 ghcLibParserIncludeDirs _forDepends ghcFlavor =
   join
     [ case ghcSeries ghcFlavor of
-        series | series >= GHC_9_10 ->
+        series | series >= GHC_9_12 ->
                  [ "rts/include"
                  , "rts/include/stg"
 #if __GLASGOW_HASKELL__ == 910 || __GLASGOW_HASKELL__ == 908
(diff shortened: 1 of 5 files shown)


shayne-fletcher:
@dylant-da do you have an account on hackage? can i have the details (to my gmail) and we can get you listed you as a maintainer?

dylant-da:
I should have an account on hackage, but I haven't used it in a while, so it might have gotten culled. Let me check.

ghc-lib-gen/src/Ghclibgen.hs
```diff
@@ -87,7 +87,8 @@ ghcLibParserIncludeDirs _forDepends ghcFlavor =
         series | series >= GHC_9_10 ->
                  [ "rts/include"
                  , "rts/include/stg"
-#if __GLASGOW_HASKELL__ == 910 || __GLASGOW_HASKELL__ == 908
+#if __GLASGOW_HASKELL__ == 910
```
dylant-da, line 90: Is there a reason we're dropping 908 here?

dylant-da (approved):
LGTM, but I'd like to confirm a question before merging

ghc-lib-gen/src/Ghclibgen.hs
shayne-fletcher: good catch. have pushed again with "the right" fix. it's busy work. the hackage linter won't allow us to upload a package with an empty directory so this has to be guarded against. it used to allow this but it became stricter.

dylant-da: :+1:

PR #611, opened 2025-09-15
thread 'allow-newer' flags

shayne-fletcher:
**rebase on master after https://github.com/digital-asset/ghc-lib/pull/609 lands**

thread `--allow-newer` flags through to nested cabal commands

this allows us to new alpha compiler releases e.g. build ghc_lib-flavor ghc-9.14.1 w/ build compiler ghc-9.14.0.20250908 (zsh)
```
rm cabal.project&&typeset -a GHC_LIB_ALLOW_NEWER=(
  --allow-newer=hashable:containers
  --allow-newer=hashable:ghc-bignum
  --allow-newer=hashable:base
  --allow-newer=unordered-containers:template-haskell
  --allow-newer=primitive:base
  --allow-newer=tagged:template-haskell
  )
cabal run exe:ghc-lib-build-tool "${GHC_LIB_ALLOW_NEWER[@]}" -- --ghc-flavor ghc-9.14.1 --no-checkout "${GHC_LIB_ALLOW_NEWER[@]}"
```

PR #612, opened 2025-09-15
toolchain issue

shayne-fletcher:
**rebase on master after https://github.com/digital-asset/ghc-lib/pull/611 lands**

toolchain is now a boot library since https://gitlab.haskell.org/ghc/ghc/-/commit/e07b031adfc542945e93d7e158eada88443972f7

this fixes the broken build of HEAD since fri


dylant-da (approved):
Looks good - I assume the new `text` dependency stems from adding ghc-toolchain as a boot library?

shayne-fletcher:
> Looks good - I assume the new `text` dependency stems from adding ghc-toolchain as a boot library?

yes

> Looks like we need to rebase before we can run CI.

see my note below. please take care of it.

PR #618, opened 2025-10-09
new arg to getOptions

shayne-fletcher:
fixes build breakage since 10/8/2025 due to 129ce32d3a4f2601f5c2d1bbaeeb2fbaf8dc9ef9


dylant-da:
I've merged this to get us unblocked, I'm not sure how much you like the nested CPP `#if` statements - we can fix this in a future cleanup PR.

PR #627, opened 2026-01-19
use system libffi

shayne-fletcher:
[hlint-from-scratch](https://github.com/shayne-fletcher/hlint-from-scratch) builds discovered the ghc-lib build was broken by https://gitlab.haskell.org/ghc/ghc/-/commit/0f53ccc6ef7560ec72b3b489b533a8ae9c6c86c1 landed 5 days ago. this fixes it.

this hasn't surfaced in the ghc-lib build because it hasn't run due to needing re-enablement for more than 30 days as pointed out in https://github.com/digital-asset/ghc-lib/issues/626

cc @nycnewman @dylant-da

### DIFF
--- a/.github/workflows/ghc-lib-ghc-HEAD-ghc-9.12.1.yml
+++ b/.github/workflows/ghc-lib-ghc-HEAD-ghc-9.12.1.yml
@@ -20,17 +20,21 @@ jobs:
       - name: Install build tools (macOS)
         run: brew install automake
         if: matrix.os == 'macos'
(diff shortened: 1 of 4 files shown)


dylant-da (approved):
Looks good to me, thank you for this - sorry for the delay, we've been in the middle of a migration between repos that is a bit all-hands-on-deck.

I've flagged up with @paulbrauner-da so that we can explore options to make the team more responsive to your PRs, and not letting things fall to the wayside that need to be addressed.

PR #629, opened 2026-09-09
allow happy-2.2

juhp:
untested


juhp:
I have hereby read the Digital Asset CLA and agree to its terms

